PR #52, opened 2020-03-17
(WIP) Register fill PDF forms

jlev:
Using pdfrw for pure python awesomeness


app/register/tasks.py
```diff
@@ -1,11 +1,62 @@
...
-def process_registration_submission(registration_pk, state_id):
+def process_registration_submission(registration_pk, state_id_number, is_18_or_over):
     registration = Registration.objects.get(pk=registration_pk)
+
+    # TODO, select template based on state
+    template_path = "register/templates/pdf/eac-nvra.pdf"
+
+    # open file objects
+    template_pdf = open(template_path, "rb")
+    filled_pdf = io.BytesIO()
```
jlev: using bytesIO instead of NamedTemporaryFile, so we are operating all in memory

app/register/tasks.py
```diff
@@ -50,6 +51,21 @@ def process_registration_submission(registration_pk, state_id_number, is_18_or_o
         suffix_field = registration.previous_suffix.replace(".", "").lower()
         registration_data[f"previous_suffix_{suffix_field}"] = True
 
+    # get mailto address from StateInformation
+    # later this will be more complicated...
+    try:
+        state_mailto_address = StateInformation.objects.get(
```
jlev: Nick, let me know if you'd prefer to do the StateInformation lookup somewhere else than in the task

app/register/tasks.py
```diff
@@ -1,11 +1,38 @@
...
+
+    # TODO, select template based on state
+    template_path = 'register/templates/pdf/eac-nvra.pdf'
+
+    # open file objects
+    template_pdf = open(template_path,'rb')
+    filled_pdf = tempfile.NamedTemporaryFile()
+
+    # convert Registration to dict
+    registration_data = registration.__dict__
```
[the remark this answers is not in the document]
nickcatal: I worry about not explicitly spelling out the fields somewhere. I do that in https://github.com/vote/turnout/blob/master/app/register/serializers.py#L18-L55

Maybe we spell it out in a `consts.py` file or something then both `serializers.py` and `tasks.py` import from that?

app/register/tasks.py
nickcatal: 👍

app/register/tasks.py
```diff
@@ -1,11 +1,76 @@
...
 from .models import Registration
 
+logger = logging.getLogger("register")
+
 
 @shared_task
 @statsd.timed("turnout.register.process_registration_submission")
-def process_registration_submission(registration_pk, state_id):
+def process_registration_submission(registration_pk, state_id_number, is_18_or_over):
     registration = Registration.objects.get(pk=registration_pk)
```
nickcatal: Can you change this to `registration = Registration.objects.select_related().get(pk=registration_pk)`?

app/register/tasks.py
nickcatal: I mean, presumably we'd want to have this here? Can you make this `Stateinformation.objects.only("text").get()` though?

app/register/validators.py
```diff
@@ -1,7 +1,5 @@
 import logging
```
nickcatal: Just a flag that I ditched this file

app/register/tasks.py
jlev: Hm, when I do that I get this error:
`django.db.models.query_utils.InvalidQuery: Field StateInformation.field_type cannot be both deferred and traversed using select_related at the same time.`

jlev: ah, had to do `Stateinformation.objects.only("field_type", "text").get()` because the manager is also doing `select_related("field_type")`

PR #53, opened 2020-03-18
Outgoing email backend

nickcatal:
In order to speed up email sending, this adapts a Sendgrid django email backend to use redis and celery.

[ch313]


app/mailer/backend.py
```diff
@@ -0,0 +1,36 @@
...
+from django.conf import settings
+from django.core.cache import cache
+from sendgrid_backend import SendgridBackend
+
+from .tasks import send_sendgrid_mail
+
+
+def push_to_sendgrid_task(message):
+    key = f"email-{uuid.uuid4()}"
+    cache.set(key, message, settings.EMAIL_SEND_TIMEOUT + 600)
```
jlev: What's the purpose of adding 600 seconds to EMAIL_SEND_TIMEOUT here? Do we expect that the mail should be sent within 10 minutes? Should that be configurable?

app/mailer/tests/test_tasks.py
```diff
@@ -0,0 +1,53 @@
+import pytest
+
+from mailer.tasks import send_sendgrid_mail
+
+TEST_MAIL = {
+    "from": {"name": "Barack", "email": "barack@turnout.local"},
```
jlev: I'd prefer if we didn't use real people for test coverage.

app/mailer/backend.py
nickcatal: An arbitrary padding so that there is never a case where the task runs after the item has expired from the cache. Can likely be much less than 10 minutes.

PR #60, opened 2020-03-27
Add latest params for local execution of turnout

nickcatal:
We've added a bunch of extra environment variables, including AWS ones, that are needed so we can execute turnout locally using production variables. These are now in remote_run

I've also moved some variables out of ECS task definitions and into AWS systems manager, so that we can more easily work with them off of ECS.


nickcatal:
FYI @jlev

PR #61, opened 2020-03-28
(WIP) Register PDF and Download Functionality

nickcatal:
This generates a registration PDF and lets you download that PDF.

(New PR because #52 was trying to merge into a feature branch and this merges into master)


app/register/generateform.py
```diff
@@ -0,0 +1,132 @@
...
+from common import enums
+from common.analytics import statsd
+from common.pdf import fill_form, join_files
+from election.models import StateInformation
+from storage.models import StorageItem
+
+logger = logging.getLogger("register")
+
+
+TEMPLATE_PATH = "register/templates/pdf/eac-nvra.pdf"
```
jlev: only issue with this is these will be variable as soon as we add absentee forms

app/register/generateform.py
```diff
@@ -0,0 +1,132 @@
...
+COVER_SHEET_PATH = "register/templates/pdf/cover.pdf"
```
jlev: this may also be variable in the future, depending on user state and absentee rules

PR #64, opened 2020-03-30
Force user's browser to download files

nickcatal:
If a user's browser opens a PDF or image in the browser, then they refresh or share the URL in their browser's URL bar, it will result in an ugly S3 "this link has expired" error.

In addition, in our effort to dedupe files on S3 we choose pretty ugly custom-per-file URLs. To keep things clean on a user's computer, we should have users download the file with the original filename.


nickcatal:
> works for me
> for reference, where is the `original_name` set?

I have to set it inside `get_available_name` so that it's available in `get_object_parameters`. There is a lot of work overriding private methods in django-storages so you could pass the original name from `get_available_name` to `get_object_parameters`

PR #65, opened 2020-03-30
API Docs

jlev:
- management command to generate openapi docs, which we use for vote/turnout-apidocs
- improved experience at `v1/fields/{{slug}}` to include state values


jlev:
force pushed to remove squashed commits after rebasing to master

jlev:
merging, so this can go into our next release

PR #66, opened 2020-03-31
Partner Association

nickcatal:
You can now add ?partner=uuid-of-partner at the end of the verifier and registration API endpoints to associate that verification/registration to an individual partner.

If no partner is identified, it will default to the first client.


app/verifier/tests/test_api_views.py
```diff
@@ -122,6 +124,66 @@ def test_lookup_object_created(requests_mock):
...
+    assert response.json() == {"registered": False}
+
+    lookup = Lookup.objects.first()
+    assert lookup.partner == first_partner
+
+
+@pytest.mark.django_db
+def test_custom_partner(requests_mock):
+    client = APIClient()
+    targetsmart_call = requests_mock.register_uri(
```
jlev: is mocked TargetSmart call necessary in this test? is it just to ensure that the VALID_LOOKUP returns 200?

app/multi_tenant/mixins_serializers.py
```diff
@@ -0,0 +1,34 @@
...
+        request = self.context.get("request")
+        if request and "partner" in request.GET:
+            try:
+                partner = Client.objects.get(pk=request.GET["partner"])
+            except (Client.DoesNotExist, ValueError, TypeError):
+                logger.info(f'Invalid partner {request.GET["partner"]}')
+                statsd.increment("turnout.multi_tenant.unknown_partner_request")
+                partner = Client.objects.first()
+        else:
+            partner = Client.objects.first()
```
jlev: this is fine, as long as there is a default Client created for us, and on application installation for other users of this code

nickcatal: There is, it happens in the 2nd migration of `multi_tenant`

app/verifier/tests/test_api_views.py
nickcatal: It's necessary so that the full serializer runs

PR #67, opened 2020-03-31
New Registration Email Notification

nickcatal:
Code to generate the outgoing email notification after a new registration is created


app/register/templates/register/email/file_notification.html
```diff
@@ -0,0 +1,55 @@
...
+            </center>
+        </td>
+    </tr>
+</table>
+<!--[if gte mso 9]>
+<br>
+<![endif]-->
+
+Once it's printed, follow the instructions on the form to sign, fold and mail it to this address:<br><br>
+{{ state_info.registration_nvrf_submission_address|markdownify }}<br><br>
```
jlev, line 37: can we make sure that this filter produces line breaks correctly?
each submission address line should end with a trailing space

PR #68, opened 2020-04-01
SMS Opt In boolean

jlev:
Required for CTIA shortcode compliance.
Includes migration.


nickcatal:
In the tests file can you check that the field is set in the `test_lookup_object_created` test?

PR #77, opened 2020-04-05
Track registration status

jlev:
Simple registration status api_view for [ch492]
Not as DRY as I'd like, but I'm having issues with DRF and enum fields


app/register/tasks.py
```diff
@@ -22,4 +23,7 @@ def send_registration_notification(registration_pk: str) -> None:
     from .notification import trigger_notification
 
     registration = Registration.objects.select_related().get(pk=registration_pk)
+    registration.status = TurnoutRegistrationStatus.PDF_SENT
+    registration.save()
```
nickcatal, line 27: We interact with `Registration` objects so weirdly and asynchronously, and these models have so many fields, I've been selectively changing fields with `registration.save(update_fields=["status"])` throughout these tasks, instead of just `.save()`

app/register/serializers.py
```diff
@@ -2,10 +2,12 @@
 import logging
 
 from enumfields.drf.serializers import EnumSupportSerializerMixin
+from enumfields.drf.fields import EnumField
```
nickcatal, line 5: Can you run `make lint` to sort these imports?

app/register/api_views.py
```diff
@@ -70,3 +70,16 @@ def process_serializer(self, serializer):
...
+    model = Registration
+    serializer_class = RegistrationSerializer
+    queryset = Registration.objects.all()
+
+    def update(self, request, *args, **kwargs):
+        instance = self.get_object()
+        serializer = StatusSerializer(instance, data=request.data, partial=True)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save(**serializer.validated_data)
+            return Response({'status': instance.status.value})
```
nickcatal, line 85: Is this `update()` even necessary? `UpdateModelMixin` should, in theory, have everything you need. Right?

app/register/api_views.py
```diff
@@ -70,3 +70,16 @@ def process_serializer(self, serializer):
             )
 
         return Response(response)
+
+class StatusViewSet(UpdateModelMixin, GenericViewSet):
+    permission_classes = [AllowAny]
+    model = Registration
+    serializer_class = RegistrationSerializer
+    queryset = Registration.objects.all()
```
nickcatal, line 78: `Registration.objects.filter(status=TurnoutRegistrationStatus.INCOMPLETE)`?

app/register/api_views.py
jlev: good point, changed the serializer for the ViewSet, and removed the `update` function

app/register/serializers.py
```diff
@@ -111,3 +111,11 @@ class StatusSerializer(EnumSupportSerializerMixin, serializers.ModelSerializer):
     class Meta:
         model = Registration
         fields = ("status",)
+
+    def validate_status(self, value):
+        if not value == enums.TurnoutRegistrationStatus.OVR_REFERRED:
+            raise serializers.ValidationError(
+                f"Registration status can only be {enums.TurnoutRegistrationStatus.OVR_REFERRED}"
```
nickcatal, line 118: `f"Registration status can only be {enums.TurnoutRegistrationStatus.OVR_REFERRED.value}"`

Right now it just shows the label, not the expected input

PR #81, opened 2020-04-06
Order registration by `-created_at`

nickcatal:
Due to a mix up in #79 this update was not merged into master. Redoing the work so that it can be merged into master.

cc @jlev


jlev:
👍

PR #83, opened 2020-04-06
(WIP) Add event tracking for Registrations and Lookups

nickcatal:
For analytics purposes we'd like to be able to track events such as official website visits. In order to make this reproducible we can use a new OneToOneField which links to an `Action` object that has events that foreign key to it. It's pretty easy to add this object to both Registrations and Lookups, then return the action id.

After the action id is acquired, you can POST to `/v1/event/track/` the action ID as `action` and the event type as `event_type`

Currently the only event type allowed is `OToolVisit` (or official state tool visit) although this can be expanded on by updating the enums.


app/event_tracking/migrations/0001_initial.py
```diff
@@ -0,0 +1,32 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Event',
+            fields=[
+                ('created_at', models.DateTimeField(auto_now_add=True)),
+                ('modified_at', models.DateTimeField(auto_now=True)),
+                ('uuid', django_smalluuid.models.SmallUUIDField(default=django_smalluuid.models.UUIDDefault(), editable=False, primary_key=True, serialize=False, unique=True)),
+                ('event_type', enumfields.fields.EnumField(enum=common.enums.EventType, max_length=10)),
```
jlev, line 25: we may want more than 10 characters here for event types

app/common/enums.py
```diff
@@ -99,3 +99,7 @@ class FileType(Enum):
 
     class Labels:
         REGISTRATION_FORM = "Registration Form"
+
+
+class EventType(Enum):
+    OFFICIAL_TOOL_VISIT = "OToolVisit"
```
jlev, line 105: let's talk through what event types we're going to need, and name them consistently

app/event_tracking/migrations/0001_initial.py
nickcatal: Changed to 100

PR #85, opened 2020-04-08
Make all "Action" serializers consistent

nickcatal:
We were repeating ourselves quite a bit inside serializers that all created new Actions. This streamlines things a bit and makes it easier to flesh out Action functionality down the line without rewriting a bunch of serializers.


app/action/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, instance=None, data=empty, **kwargs) -> None:
+        self.incomplete = kwargs.pop("incomplete", False)
+        super().__init__(instance=instance, data=data, **kwargs)
+
+    def create(self, validated_data: Dict[(str, Any)]) -> "Model":
+        validated_data["action"] = Action.objects.create()
+        return super().create(validated_data)
+
+    def get_field_names(
+        self, declared_fields: Mapping[(str, serializers.Field)], info: "FieldInfo"
```
jlev, line 31: Do linting rules suggest putting these on newlines? I know types is the new python way, but I find this hard to parse.

nickcatal: It's what the linter says

PR #87, opened 2020-04-09
Official Election Information: cache, sync, api

nickcatal:
[ch507]


jlev:
updated state endpoint to be nested
@nickcatal can you review?

app/official/serializers.py
```diff
@@ -0,0 +1,35 @@
...
+class RegionNameSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Region
+        fields = ("name", "external_id")
+
+
+class AddressSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Address
+        fields = (
```
nickcatal, line 14: Random request, can you put these inside lists instead of tuples?

app/official/migrations/0001_squashed.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class Migration(migrations.Migration):
+
+    replaces = [('official', '0001_initial'), ('official', '0002_address_office'), ('official', '0003_auto_20200410_2042'), ('official', '0004_auto_20200410_2047'), ('official', '0005_address_zipcode')]
+
+    initial = True
+
+    dependencies = [
+        ('election', '0004_updatenotificationwebhook'),
```
nickcatal, line 16: To keep things clean, can you remove all these migrations and create a new `0001_initial`? Then change this line to 
`('election', '0001_initial')`

Will probably have to run `python manage.py migrate official zero` locally before you do this

app/official/tasks.py
```diff
@@ -0,0 +1,8 @@
+from celery import shared_task
+
+
+@shared_task
+def sync_usvotefoundation():
```
nickcatal, line 5: We should add a cron for this. I can add it later though.

app/official/models.py
```diff
@@ -0,0 +1,63 @@
...
+    fax = PhoneNumberField(null=True)
+
+    is_physical = models.BooleanField(default=False)
+    is_regular_mail = models.BooleanField(default=False)
+
+    process_domestic_registrations = models.BooleanField(default=False)
+    process_absentee_requests = models.BooleanField(default=False)
+    process_absentee_ballots = models.BooleanField(default=False)
+    process_overseas_requests = models.BooleanField(default=False)
+    process_overseas_ballots = models.BooleanField(default=False)
```
nickcatal, line 60: These `BooleanField`s all need `null=True` as well, in addition to `default=False`

app/official/tasks.py
jlev: do you mind doing this? defer to your knowledge about the production docker setup

PR #89, opened 2020-04-13
Allow custom partner slugs

nickcatal:
If a partner wishes to link an individual signup to their organization they need to pass in a specific `partner` parameter. Previously this was the SmallUUID primary key of the `Client`, but this had a few problems. (1) the UUID was ugly and meaningless, (2) it could never change. By allowing custom "slugs" for partners, partners can have use meaningful names in their links, and by making it a foreign key we can allow partners to have multiple different slugs instead of forcing them to rename existing slugs (and thus break old links.)

Slugs are managed in the `Client` section of the Django Admin.

This also exposes a new endpoint, `/v1/partner/id/slug/`, which will return the name of the partner. This could be useful for instances where we need to disclose the name of the partner associated with any one signup so that the user can be opted into that partner's email or SMS list.


nickcatal:
I'm not thinking multiple slugs will be for partners individual tracking as much as an insurance policy for us down the line if a partner wants to change slugs but doesn't want to break the links they have in the wild. We'll intentionally not add additional slugs.

PR #90, opened 2020-04-14
(WIP) Absentee API Endpoint

nickcatal:
Since the absentee and register endpoints are so similar, in order to prevent duplicate code this refactors the register codebase and puts it into the `action` app.

I forgot to actually commit the absentee model in #86, so that's included in this as well.

I generalized the action status enum.


app/absentee/models.py
```diff
@@ -0,0 +1,61 @@
...
+from common.utils.models import TimestampModel, TrackingModel, UUIDModel
+from common.validators import zip_validator
+from multi_tenant.mixins_models import PartnerModel
+
+
+class BallotRequest(
+    ActionModel, PartnerModel, TrackingModel, UUIDModel, TimestampModel
+):
+    first_name = models.TextField(null=True)
+    middle_name = models.TextField(null=True)
```
jlev, line 16: middle_name not required, should be blank=True

app/absentee/models.py
```diff
@@ -0,0 +1,61 @@
...
+    last_name = models.TextField(null=True)
+    suffix = models.TextField(null=True)
```
jlev, line 18: suffix not required, should be blank=True

app/absentee/models.py
```diff
@@ -0,0 +1,61 @@
...
+    date_of_birth = models.DateField(null=True)
+    email = models.EmailField(null=True)
+    phone = PhoneNumberField(null=True)
```
jlev, line 21: phone not required, should be blank=True

PR #91, opened 2020-04-14
PDF generation for absentee ballot request

benweissmann:
- Small refactor to PDF generation helpers
- Add a module to create absentee ballot requests

This is still WIP -- there's a few outstanding TODOs, and it's not hooked into the API. You can test it out with `./manage.py testabsenteereq`.


app/absentee/generateform.py
```diff
@@ -0,0 +1,103 @@
...
+    ).fill(form_data)
+
+
+def prepare_formdata(region_external_id: str, state_code: str) -> Dict[str, Any]:
+    form_data: Dict[str, Any] = {}
+
+    # find the mailing address
+
+    mailing_addresses = Address.objects.filter(
+        office__region__external_id=region_external_id, process_absentee_requests=True
```
jlev, line 43: also need a filter here for is_regular_mail=True
some offices have two addresses, one for physical location and one for mailing

app/absentee/generateform.py
```diff
@@ -0,0 +1,103 @@
...
+    try:
+        state_mail_deadline = (
+            StateInformation.objects.only("field_type", "text")
+            .get(state=state_code, field_type__slug="vbm_deadline_mail")
+            .text.lower()
+        )
+
+        # TODO: are the deadlines mail-by or arrive-by? Putting arrive-by here
+        # becuase that's more conservative
+        state_deadline = f"Your form must arrive by {state_mail_deadline}."
```
jlev, line 77: deadlines will say in their text whether they are postmarked by or arrive by
working on cleaning that up in the spreadsheet before we import to API

benweissmann: I'm not sure that's the case for NBM deadline (or maybe I'm looking at the wrong field?): https://admin.voteamerica.com/manage/election/fieldinformationtype/vbm_deadline_mail/

jlev: You're right, the current field values for `vbm_deadline_mail` don't indicate received or postmarked. I'll flag this with the research team.

The voted ballot deadline does differ between states, but haven't imported that yet.

app/absentee/management/commands/testabsenteereq.py
```diff
@@ -0,0 +1,21 @@
+from django.core.management.base import BaseCommand
```
nickcatal, line 1: You're going to want to avoid management commands in general for django. They're hard to run and we would add complexity to running all the project's tasks locally and on our CI system.

You can use pytest inside of `absentee/test/test_generateform.py`

app/register/generateform.py
```diff
@@ -107,7 +87,7 @@ def extract_formdata(registration, state_id_number, is_18_or_over):
         state_mail_deadline = (
             StateInformation.objects.only("field_type", "text")
             .get(
-                state=registration.state, field_type__slug="registration_deadline_mail",
+                state=registration.state, field_type__slug="registration_deadline_mail"
```
nickcatal, line 96: This might be here due to linting? `make lint` may end up re-adding this

app/common/pdf.py
```diff
@@ -101,3 +109,42 @@ def join_files(input_files, output_file):
...
     output.write(output_file)
+
+
+class PDFTemplate:
+    """
+    Constructs a template out a set of input PDFs with fillable AcroForms.
+    """
+
+    def __init__(self, template_files: List[str]):
+        self._template_files = template_files
```
nickcatal, line 22: There's probably no reason to use an underscore here and keep this private. `self.template_files = template_files`

app/common/pdf.py
```diff
@@ -1,103 +1,76 @@
...
+                    pdf_path=template_file.path,
+                    datas=data,
+                    out_file=filled_template.name,
+                    flatten=False,
+                )
+
+                filled_templates.append(filled_template.name)
+
+            # Join the filled forms
+            pypdftk.concat(files=filled_templates, out_file=final_pdf.name)
```
nickcatal, line 67: I worry that we won't get a ton of telemetry about PDF generation, since this is technically an external service outside our python process. One option here is to add a custom datadog service for pypdftk, similar to what we do to instrument sendgrid here https://github.com/vote/turnout/blob/master/app/mailer/tasks.py#L21-L22

you can do that with something like this

```python
from common.apm import tracer
for template_file in template_files:
    # stuff above pdftk here
    with tracer.trace("pdftk.form_fill", service="pdftk"):
        pdftk.fill_form(stuff)
    # stuff below form fill here

with tracer.trace("pdftk.concat", service="pdftk"):
    pypdftk.concat(files=filled_templates, out_file=final_pdf.name)
```

app/common/pdf.py
```diff
@@ -1,103 +1,76 @@
...
+
+        except:
+            final_pdf.close()
```
nickcatal, line 70: We want to keep a count of the number of exceptions here using datadog

```python
from common.analytics import statsd
statsd.increment("turnout.pdf.pdftk_exception")
```

app/register/generateform.py
```diff
@@ -125,13 +111,13 @@ def extract_formdata(registration, state_id_number, is_18_or_over):
...
     form_data = extract_formdata(registration, state_id_number, is_18_or_over)
 
-    filled_pdf = generate_pdf(form_data)
-    item = StorageItem(
-        app=enums.FileType.REGISTRATION_FORM,
-        email=registration.email,
-        partner=registration.partner,
-    )
-    item.file.save(generate_name(registration), File(filled_pdf), True)
+    with generate_pdf(form_data) as filled_pdf:
```
nickcatal, line 114: I'm blanking here regarding how file handling is handled, is the reason we're using `with` because it closes the file?

app/requirements.txt
```diff
@@ -51,4 +51,5 @@ netifaces==0.10.*
 pdfrw==0.4
```
nickcatal, line 51: Do we need this requirement if we're moving to pdftk?

Dockerfile-server-dev
```diff
@@ -1,6 +1,7 @@
 FROM python:3.7-buster
-RUN apt-get update && apt-get install -y \
-    wait-for-it && apt-get clean
+RUN apt-get update && \
+    apt-get install -y wait-for-it pdftk-java && \
+    apt-get clean
```
nickcatal, line 4: Yay! This is perfect

scripts/docker_build_step1.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 curl -sL https://deb.nodesource.com/setup_12.x | bash -
-apt-get install -y nodejs
+apt-get install -y nodejs pdftk-java
```
nickcatal, line 4: Exactly where this needs to be. Great!

Makefile
```diff
@@ -22,6 +22,8 @@ shell:
 clientshell:
 	docker-compose exec client /bin/bash
 
+workershell:
+	docker-compose exec worker /bin/bash
```
nickcatal, line 26: `workershell` and `shell` should be identical, except the worker will have the celery celery process running in the background and server will have the django dev server process running in the background.

Whereas `clientshell` will enter you into a client that has npm and all the static asset requirements installed, and none of the python requirements.

So this is likely redundant.

app/absentee/management/commands/testabsenteereq.py
benweissmann: Right, this is just a temporary way for me to be able to actually run this code manually since it's not hooked up to the API -- I'll delete this whole file from this PR once I have it hooked into the API.

app/register/generateform.py
benweissmann: Weird, I don't remember removing this comma and I'm running Black in my editor too -- this looks like a case where Black is unopinionated and won't add or remove it. I'll put the comma back.

app/register/generateform.py
benweissmann: Yup, it's equivalent to having an `ensure` block that closes the file even in the case of an exception.

app/requirements.txt
benweissmann: Will remove for now -- we'll probably need it back when we do signatures since that's a pain with pdftk. For context, Josh and I have been talking about pdftk vs. pdfrw and it's still up in the air whether we want to go with pdftk; let's talk more today and make a call.

Makefile
benweissmann: I was using this to get into the worker specifically to debug pdftk -- by passing `delete=False` when constructing the `NamedTemporaryFile`s you can have python leave the intermediate PDFs in `/tmp` so you inspect them.

That said, this is really just a workaround to be able to manually re-run the `pdftk` binary and see the stdout/stderr because pypdftk doesn't do a good job of reporting them; if we change pypdftk to do a better job of logging then we probably don't need this (or can at least just let people `docker-compose exec` manually)

app/absentee/generateform.py
```diff
@@ -0,0 +1,120 @@
...
+from election.models import StateInformation
+from storage.models import StorageItem
+
+from .contactinfo import get_absentee_contact_info
+from .models import BallotRequest
+from .tasks import send_ballotrequest_notification
+
+logger = logging.getLogger("absentee")
+
+COVER_SHEET_PATH = "absentee/templates/pdf/cover.pdf"
```
jlev, line 19: at some point we may have state specific coversheets, if there are big warnings we need to apply that are not doable in the standard text fields

app/absentee/generateform.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def generate_name(state_code: str, last_name: str):
+    """
+    Generates a name for the PDF
+    """
+    filename = slugify(f"{state_code} {last_name} ballotrequest").lower()
```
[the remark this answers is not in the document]
jlev: > code looks good
> tested locally with PDF generated for MA
> CA had error:
> 
> ```
> worker_1    | Traceback (most recent call last):
> worker_1    |   File "/usr/local/lib/python3.7/site-packages/celery/app/trace.py", line 385, in trace_task
> worker_1    |     R = retval = fun(*args, **kwargs)
> worker_1    |   File "/usr/local/lib/python3.7/site-packages/celery/app/trace.py", line 650, in __protected_call__
> worker_1    |     return self.run(*args, **kwargs)
> worker_1    |   File "/usr/local/lib/python3.7/site-packages/datadog/dogstatsd/context.py", line 56, in wrapped
> worker_1    |     return func(*args, **kwargs)
> worker_1    |   File "/app/absentee/tasks.py", line 16, in process_ballotrequest_submission
> worker_1    |     process_ballot_request(ballot_request)
> worker_1    |   File "/app/absentee/generateform.py", line 103, in process_ballot_request
> worker_1    |     with generate_pdf(form_data, ballot_request.state.code) as filled_pdf:
> worker_1    |   File "/usr/local/lib/python3.7/site-packages/datadog/dogstatsd/context.py", line 56, in wrapped
> worker_1    |     return func(*args, **kwargs)
> worker_1    |   File "/app/absentee/generateform.py", line 95, in generate_pdf
> worker_1    |     ).fill(form_data)
> worker_1    |   File "/app/common/pdf/pdftemplate.py", line 70, in fill
> worker_1    |     pypdftk.concat(files=filled_templates, out_file=final_pdf.name)
> worker_1    |   File "/usr/local/lib/python3.7/site-packages/ddtrace/tracer.py", line 620, in func_wrapper
> worker_1    |     return f(*args, **kwargs)
> worker_1    |   File "/app/common/pdf/pypdftk.py", line 177, in concat
> worker_1    |     run_command(args)
> worker_1    |   File "/app/common/pdf/pypdftk.py", line 44, in run_command
> worker_1    |     result.check_returncode()
> worker_1    |   File "/usr/local/lib/python3.7/subprocess.py", line 444, in check_returncode
> worker_1    |     self.stderr)
> worker_1    | subprocess.CalledProcessError: Command '['/usr/bin/pdftk', '/tmp/tmpnvpb3ohf', 'absentee/templates/pdf/states/CA.pdf', 'cat', 'output', '/tmp/tmpao3r6rn9']'
> ```
> 
> Seems like we had different places we were expecting the PDF templates to live.

resolved in the merge

PR #93, opened 2020-04-15
Minor version bumps to resolve security alerts

benweissmann:
This resolves the security alerts on the Github repo as well as a bunch of minor ones reported by `npm audit`. They mostly seem to be from development dependencies (transitive webpack dependencies) so they're not super relevant but I figured that cleaning them up would make us more likely to notice any real alerts in the future.


jlev:
Good idea, those security warnings are frequent enough that I basically ignore them...

PR #98, opened 2020-04-17
Change EnumFields from CharFields to TextFields

nickcatal:
Postgres treats Charfields and TextFields identically, except Charfields can have a character limit and TextFields do not. We have no reason to have a character limit on our EnumFields, so we can change them to just be textfields.

[ch611]


jlev:
👍

PR #101, opened 2020-04-17
add statewatch API endpoint

liewegas:
Expose results of (modified) urlwatch scraper via REST.


jlev:
Sage, does this assume that the URLWatch is running on another process and inserting the checks into the database? We may want to run this in a separate app, but let's talk about it next week. We have a lot of other PRs going on to get our absentee voter tool operational by the end of the month.

PR #105, opened 2020-04-24
Verify check Alloy api

jlev:
needs more tests, but works


jlev:
@nickcatal let's talk about which characters should actually be allowed in the targetsmart request. their docs are insufficiently precise.

jlev:
tests improved. let's get this merged tomorrow, so we can reduce our dependency on one single API for verify

PR #106, opened 2020-04-26
Cloudflare-based API Acceleration

nickcatal:
Thanks to Cloudflare we're now able to accelerate our API by having our CDN cache API responses that do not change often (such as state election information) and only have requests for this content come to the origin when there is a change.

This is done using the `Cache-Tag` header. By categorizing request types by tags such as environment, django model, app, object primary key, etc then including that tag in a header in every outgoing request we can purge entire categories of requests that may have involved returning details about that data.

The first test of this will be using the state information API.

Cloudflare's documentation about this functionality: https://support.cloudflare.com/hc/en-us/articles/200169246


app/cdn/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    final_tags = [generate_tag(tag) for tag in tags]
+    extra = {
+        "tags": final_tags,
+    }
+    logger.info("Cache Tag Purge %(tags)s", extra, extra=extra)
+    if not settings.CLOUDFLARE_ENABLED:
+        logger.info("Cloudflare Disabled")
+        return
+    cf = CloudFlare.CloudFlare(token=settings.CLOUDFLARE_TOKEN)
+    cf.zones.purge_cache.delete(
```
benweissmann, line 28: How do we want to handle failures here? Should we defer this to a queued task so we get retry behavior?

app/cdn/utils.py
```diff
@@ -0,0 +1,34 @@
...
+from django.conf import settings
+from django.template.defaultfilters import slugify
+
+from common.analytics import statsd
+
+logger = logging.getLogger("cdn")
+
+
+def generate_tag(raw_tag):
+    return slugify(f"to{settings.ENV}{str(raw_tag)}".lower())
```
benweissmann, line 14: What's the meaning behind this format and the `to` prefix? Should there be some separator here so we don't end up with collisions?

app/cdn/middleware.py
```diff
@@ -0,0 +1,33 @@
...
+
+        if not hasattr(response, "cache_tags"):
+            response.cache_tags: List[str] = []
+
+        new_tags = []
+
+        new_tags.append(f"detail_{settings.CLOUD_DETAIL}")
+        new_tags.append(f"tag_{settings.TAG}")
+        new_tags.append(f"build_{settings.BUILD}")
+        new_tags.append(f"env_{settings.ENV}")
```
benweissmann, line 25: It looks like `generate_tag` already adds the environment to every tag -- do we need a specific env tag as well?

app/cdn/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@statsd.timed("turnout.cdn.purge_tags")
+def purge_cdn_tags(tags: Iterable[str]) -> None:
```
benweissmann, line 18: Should this get run after deploys? What happens if we add a new field to the API response -- would the cache serve the old response without the new field?

app/cdn/utils.py
nickcatal: Cloudflare caches are scoped to an entire domain ("zone" in their parlance), so each key needs to be unique from literally every other key on every resource on every subdomain of that domain, or your apps will start expiring keys from other resources on the same domain.

Keys should also be relatively short. There is no explicit length limit on them from Cloudflare's side, but requests to say the view that lists all our states has almost 60 tags in the response header. So every extra character when making the key unique is multiplied by 60 on those requests.

app/cdn/middleware.py
nickcatal: If we ever want to purge every cache on an entire environment we'll need a separate tag.

app/cdn/utils.py
nickcatal: Every time a new field is added or a field changes the cache is purged thanks to [django signals](https://docs.djangoproject.com/en/2.2/topics/signals/)

Every update to the information in the django admin/rest API/management interface will trigger an immediate cache expiration.

The downside here is that new requests are quite resource intensive, so the next request to the endpoint will be somewhat slow. But the upside is that because we're expiring only keys that pertain to an individual piece of information, we won't be purging every endpoint of the API.

One Q here is if we want this request to be done using a task, but I've found that purge API requests to Clouflare are really fast and you can purge multiple keys with 1 request.

Here's the implementation of signals:
https://github.com/vote/turnout/pull/106/files#diff-ccec28b69fb03689172e1fd25c4e2cb6R128-R140

benweissmann: Right, but what if a code change alters the response? For example, let's say that we decided to at a `created_at` field to responses in addition to the `modified_at` field. Would we need to manually expire the cache after deploying?

app/cdn/utils.py
nickcatal: I dunno. Part of me wants this to just be something that, if it fails, it raises an exception vs sends it to a queue. If it raises an exception we'll have crumbs from the `logger.info` above it in sentry. Similar to what would happen if redis or a database failed. But I can likely be (easily) persuaded that we handle it all in queues.

benweissmann: My concern is around atomicity -- if the database write succeeds and then Cloudflare fails, then we end up with the cache out of sync with the database, vs. being able to roll back a database transaction. OTOH, maybe the person editing the data would just try again if they got an error? But the admin pages aren't cached (right? those are server-side-rendered HTML directly from the DB?) -- so the admin would see the fresh data and think that the edit went through, even though it's not live on the website because the cache is stale.

I guess maybe it depends on our cache lifetime? Like if we're only caching for an hour, it doesn't seem like a big deal to have the explicit invalidation fail, but if we're caching for a day or more, it seems like a bigger problem.

nickcatal: In this case if there is an exception the admin would also throw an exception and the user would see a 500 page. We can choose, with django signals, if the exception happens before the data is saved to the database (pre-save) or it's current situation of purging after the data has been committed to the database (post-save.)

The way to go here is probably have the cache clear happen pre-save, which will throw a 500 error if Cloudflare fails and cause django to bail out of the entire flow prior to actually making the change.

I was thinking we'd cache for 24 hours.

app/cdn/utils.py
nickcatal: If we ever make a forward-facing change we'd have to purge cloudflare. We don't do that a ton though, and with our red-black deployment there's no great way of purging the site and deploying the site at exactly the same moment.

Which brings up the fact that data migrations and field additions/deletions already have a few weird gotchas (like when you add a field you have to run the migration before code deploys, and make the fields all nullable, and if you want to delete a field you have to make sure that field is nullable then have the deletion migration happen after all the code that needed that field is done) so I guess manually clearing the cloudflare cache using the cloudflare UI in those cases isn't the worst thing.

We could also add some code to clear the entire hostname inside of spinnaker after a successful deploy. Spinnaker will run post-success scripts if we want.

app/cdn/utils.py
nickcatal: Worth noting that it's theoretically possible (albeit extraordinarily unlikely) that a request could come in between the time the cache is cleared and the transaction is saved to the database. Although Cloudflare has a LOT of logic to get through to expire their cache globally, so I'd say the risk that a cloudflare request dies is higher than Cloudflare somehow purges it's global cache in the few milliseconds between the request to Cloudflare's API finishing and the commit to the database.

PR #107, opened 2020-04-27
Allow stateinformation api to return as_object instead of nested list

jlev:
To make our API easier to consume by some frameworks


jlev:
🙌

PR #109, opened 2020-04-29
Add signature feature to PDFTemplate

benweissmann:
Gives PDFTemplate the ability to stamp signatures onto the PDF. For right now, you have to explicitly tell it the bounding box -- extracting the location of the bounding box from the form input is actually way harder than a thought, because the bounding box specified by the annotation item in the PDF is relative to the transformed coordinate space -- you have to go through the whole page, figure out which transforms are applied to the annotation, and then multiply the annotation location by all of those transformation matrices (i think!). I think that's doable, but we probably want to do something like store the bounding box explicitly in the repo and have a precommit hook that updates it when you change the PDF so we can adjust it by hand if it's a really wacky PDF. In any case, I think we can live with this for the MVP, and we can build out the code to automatically compute the bounding box from the PDF in a few weeks.


app/common/pdf/pdftemplate.py
```diff
@@ -22,8 +40,9 @@ class PDFTemplate:
 
     def __init__(self, template_files: List[PDFTemplateSection]):
         self.template_files = template_files
+        self.pypdftk = PyPDFTK()
 
-    def fill(self, raw_data: Dict[str, Any]) -> IO:
+    def fill(self, raw_data: Dict[str, Any], signature=None) -> IO:
```
nickcatal, line 45: Do you want to do type hinting here for `signature`?

jlev:
This approach makes sense, and I shudder to think about the matrix math that we might have to do to determine all these offsets automatically. Let's figure them out by hand for the first few states, and then write some script to compute the rest and store the output in the repo. Much better than doing on the fly, and potentially getting it wrong.

PR #111, opened 2020-04-29
Expose fields needed for disclaimers in partner embeds

nickcatal:
In order to support partners we may need to show their name as well as provide links to their terms of service and privacy policy. In addition some will need to have checkboxes that opt the user into their SMS program. Some of those checkboxes will be checked by default while others will not.

We can expose all of this in embeds.


app/multi_tenant/api_views.py
```diff
@@ -1,14 +1,28 @@
...
+        instance = self.get_object()
+        partner = instance.partner
+
+        return Response(
+            {
+                "name": partner.name,
+                "url": partner.url,
+                "terms_of_service": partner.terms_of_service,
+                "privacy_policy": partner.privacy_policy,
+                "sms_enabled": partner.sms_enabled == True,
```
jlev, line 25: this is to ensure that nulls aren't passed back?

nickcatal: Yes

PR #114, opened 2020-05-06
Default partner slug, partner SMS improvements

nickcatal:
(1) This pull request adds the ability for a partner slug to be "default", which will be useful for cases where we need to generate embed code using a partner's primary slug or where we may want to include the partner's slug in a URL in perhaps the admin
(2) This adds additional fields that we can pass to the frontend related to SMS disclaimers
(3) This allows scoping SMS opt-ins to the individual partner


app/multi_tenant/models.py
```diff
@@ -29,6 +40,25 @@ def full_email_address(self) -> str:
 class PartnerSlug(UUIDModel, TimestampModel):
     partner = models.ForeignKey(Client, on_delete=models.CASCADE)
     slug = models.SlugField(unique=True)
+    default = models.BooleanField(default=True, null=True)
+
+    def save(self, *args, **kwargs) -> None:
+        if self.default:
```
benweissmann, line 46: Does this happen in a transaction? Could we add a partial unique index to enforce this? E.g. `CREATE UNIQUE INDEX single_default ON partner_slug (partner) WHERE (default IS TRUE)`. Alternatively, if we want to enforce that there's always a default slug for each partner, maybe we want a `default_slug` field on `Partner`?

nickcatal: Refactored to store `default_slug` on partner, with a PROTECT clause, which will prevent django from deleting the default partner for a client.

PR #115, opened 2020-05-06
Add signature upload and stamping

benweissmann:
- Adds a new endpoint for "secure uploads" -- just uploading a file to us with no way to read it back or download it.
- Allows you to submit a signature along with your absentee ballot request. The PDF will be stamped with the signature if the ballot request has a signature attached, and we have metadata for the state with where to stamp the signature.
- Fixes a PDF generation bug where the returned temp file was not open in binary mode, so you'd get UTF-8 errors when uploading it
- Reworks how we resize the signature image so we get higher-quality signatures (we embed the full-quality signature file with a scaled bounding box rather than scaling down the signature to fit)


app/storage/models.py
```diff
@@ -50,3 +50,10 @@ def download_url(self):
     @property
     def reset_url(self):
         return settings.FILE_TOKEN_RESET_URL.format(item_id=self.pk)
+
+
+class SecureUploadItem(UUIDModel, TimestampModel):
+    file = models.FileField(storage=HighValueStorage())
```
nickcatal, line 56: Do we think we'll want to store an enum of the file type here, so we have some sense of what we're storing?

app/storage/api_views.py
```diff
@@ -27,3 +30,16 @@ def post(self, request):
         # TODO: Trigger email to user with new token
 
         return Response(status=status.HTTP_201_CREATED)
+
+
+class SecureUploadView(APIView):
```
nickcatal, line 35: Does it make sense to make this as an `APIView` or viewset?

If it's a viewset with a serializer then we can, in theory, relatively-simply offer a Read API later on for this object type. But if that's a hassle not a huge deal.

app/absentee/state_pdf_data.py
```diff
@@ -0,0 +1,13 @@
...
+import yaml
+
+STATE_DATA = {}
+
+for fname in glob.glob("absentee/templates/pdf/states/*.yml"):
+    state, _ = os.path.splitext(os.path.basename(fname))
+    with open(fname) as f:
+        STATE_DATA[state.upper()] = yaml.safe_load(f)
+
+print(STATE_DATA)
```
nickcatal, line 13: Unnecessary `print` here

app/storage/api_views.py
benweissmann: I think doing a viewset is going to be complicated (esp. with the new changes) -- this endpoint doesn't map super well to JSON. I think if we're going to go that direction we probably want to have this API look more like a typical JSON representation with the content as a base64-encoded string or something? But that seems uglier for the public API than just POSTing the file to the endpoint.

app/absentee/migrations/0005_ballotrequest_signature.py
```diff
@@ -0,0 +1,20 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('storage', '0004_secureuploaditem'),
+        ('absentee', '0004_enumchartotextfield'),
```
nickcatal, line 11: Why does this migration require a previous migration from the `absentee` app?

app/storage/api_views.py
nickcatal: Makes sense to me!

app/absentee/migrations/0005_ballotrequest_signature.py
benweissmann: I think django just puts the previous migration from the app in the dependency array to enforce order? I'm not super sure; this is all just autogenerated

nickcatal: Scratch that, I misread things.

PR #116, opened 2020-05-07
Absentee State Fields

jlev:
simple nullable jsonfield, no validation


jlev:
@benweissmann this isn't the full implementation we talked about, but I think it will get us to e-sign MVP more quickly

app/absentee/generateform.py
```diff
@@ -57,6 +57,8 @@ def prepare_formdata(
     form.
     """
     form_data = model_to_dict(ballot_request)
+    # insert state_fields into form_data top-level
+    form_data.update(form_data['state_fields'])
```
benweissmann, line 61: Can we make it so this doesn't overwrite fields that exist? Or add at least a validation that none of the fields of `state_fields` conflict with models fields? Otherwise this gets around any validation that we try to do on the fields.

jlev: good catch. I was trying to do it the pythonic way, but a forloop might be better instead

benweissmann:
Also @jlev from the tests it looks like your migration conflicts with the one I merged earlier today -- you'll need to merge master into your branch and then re-generate the migration

PR #119, opened 2020-05-08
Quick-and-dirty implementation of emailing LEOs for Georgia

benweissmann:
- Add a code path for emailing applications directly to LEOs. Use it for Georgia.
- Add a table for email overrides per-region
- Add an env var, `ABSENTEE_LEO_EMAIL_OVERRIDE`, that will make it always send the email to a fixed address with a note of what the LEO's email is, rather than actually emailing the LEO. In dev, this defaults to a Slack email address. We should set this env var for the dev and staging environments so we can QA without committing voter fraud


app/absentee/leo_email.py
```diff
@@ -0,0 +1,79 @@
...
+        subject=subject,
+        body=content,
+        from_email=FROM_EMAIL,
+        to=[leo_email],
+        cc=[ballot_request.email],
+    )
+
+    attachment = ballot_request.result_item.file
+    with attachment.open() as f:
+        msg.attach(attachment.name, f.read(), "application/pdf")
```
nickcatal, line 71: We probably want a cleaner name than the one that we're storing on S3

app/absentee/leo_email.py
```diff
@@ -0,0 +1,79 @@
...
+
+from common.analytics import statsd
+
+from .contactinfo import get_absentee_contact_info
+from .models import AbsenteeLeoEmailOverride, BallotRequest
+
+NOTIFICATION_TEMPLATE = "absentee/email/leo_email.html"
+
+# TODO: this should be a monitored email
+FROM_EMAIL = "noreply@voteamerica.com"
```
nickcatal, line 15: Let's store this in django settings

app/absentee/leo_email.py
```diff
@@ -0,0 +1,79 @@
...
+        leo_email = OVERRIDE_EMAIL
+        was_debug_mail = True
+
+    context = {
+        "region_name": ballot_request.region.name,
+        "voter_full_name": f"{ballot_request.first_name} {ballot_request.last_name}",
+        "voter_email": ballot_request.email,
+        "debug": was_debug_mail,
+        "leo_email": real_leo_email,
+    }
```
nickcatal, line 53: Because this email template inherits from a larger wrapper, and that wrapper expects certain fields to be in the context (like the preheader text) we should pass in context in a similar format to this: https://github.com/vote/turnout/blob/master/app/absentee/notification.py#L17-L31

So in this file you'll likely want to pass in the context like this

```python
    preheader_text = f"Here is an absentee ballot request form for {ballot_request.first_name} {ballot_request.last_name}"
    recipient = {
        "first_name": ballot_request.first_name,
        "last_name": ballot_request.last_name,
        "email": ballot_request.email,
    }
    context = {
        "ballot_request": ballot_request,
        "partner": ballot_request.partner,
        "recipient": recipient,
        "state_info": ballot_request.state.data,
        "preheader_text": preheader_text,
        "debug": was_debug_mail,
        "leo_email": real_leo_email,
    }

```

Then inside the template you'll have access to `{{ recipient.first_name }}` `{{ recipient.last_name }}`, `{{ leo_email }}` etc

benweissmann: What should I pass for recipient.first_name and recipient.last_name? This email is going to the LEO but we don't have a name for them.

benweissmann: I also took a look at the base template and it didn't look like the preheader was required -- I'm not sure there's much useful to put there so I left it out

nickcatal: You can not pass in `first_name` and `last_name`, and yeah feel free to ditch the preheader text.

nickcatal:
```python
    recipient = {
        "first_name": "",
        "last_name": "",
        "email": ballot_request.email,
    }
```

Can just be

```python
    recipient = {
        "email": ballot_request.email,
    }
```

Django templates will ignore missing keys in dictionaries if you try to reference them.

PR #120, opened 2020-05-08
Filter LEO offices by absentee capability

jlev:
Absentee ballots are processed in MI and WI at the local level, not the county. If an `?absentee` parameter is provided to `/v1/official/state/MI`, we should only include local level.

This attempts to solve that by filtering by Offices and Addresses which have the `process_absentee_requests` flag set to True, but unfortunately that is not consistently applied in the upstream data. As a fallback, in these two states we limit out regions which do not have a named municipality (eg, only county-level offices).

This is not ideal, but fixes a nasty bug quickly. We should move soon towards geocoding these offices and providing a suggestion, so users are not faced with 1900 choices.


jlev:
I remembered after I pushed that the website is actually getting this data from the serializer at the moment, when the request includes ?incomplete=true. Amended to include the MI/WI filtering there, but not the more complex office/address filtering.

We'll push to get that updated upstream...

PR #126, opened 2020-05-12
Documentation

nickcatal:
Adding mkdocs


jlev:
Added copy for static info api, tool embeds, UTM params and open code license.
Worth a review to make sure it's sufficiently detailed, without being confusing.

PR #130, opened 2020-05-12
Add extra environment vars

nickcatal:
We're going to be using Twilio soon and want to override the email that messages to LEO point to.


nickcatal:
Yeah, it should

PR #135, opened 2020-05-15
Catch exceptions from event tracking

benweissmann:
Just a quick fix for now; we still need to figure out the root cause of why we don't have a corresponding ballot request / registration record


benweissmann:
I think in this case we do want to catch all errors -- there's nothing that could go wrong with event tracking that would make us want to block users from downloading their file; I think that in this case log-and-continue is always the right option (event if the error is something else, like a connection error to the database)

PR #136, opened 2020-05-15
Update absentee forms

jlev:
[ch892]

update UT and CO
minor change to GA to check "has_mailing_address" box


jlev:
needs QA testing

jlev:
QA tested CO, UT, TX, MD, PA

GA still not showing the checkbox for `has_mailing_address`, even after I deleted and recreated it.
@benweissmann any idea what's going on here?

benweissmann:
I think the issue with the GA text box is that the value of the checkbox is "Yes" rather than "On" (we use `"On"` for True values in the code). You can find that in the "Export Options" tab of the checkbox properties:

Screen Shot 2020-05-20 at 10 01 48 AM

jlev:
tested fix for GA, and it works
also split out mailto_line_N for absentee envelope for [ch856]

PR #140, opened 2020-05-18
Add utm_term and utm_content, linting and fixture usage

nickcatal:
Added the ability to also track utm_campaign and utm_term in our app.

Also did a bit of refactoring of @liewegas's smsbot tests to use a pytest fixture


app/verifier/tests/test_api_views.py
```diff
@@ -143,9 +142,6 @@ def test_targetsmart_request_address2(requests_mock, mocker, settings):
         "1600 Penn Avenue Unit 2008, Chicago, IL 60657"
     ]
     assert targetsmart_call.last_request.qs == address2_qs
-    patched_sms_send.apply_async.assert_called_once_with(
-        args=("(312) 928-9292",), countdown=150
-    )
```
liewegas, line 148: These asserts are all getting dropped.. was that intentional?

nickcatal: Yeah, I figured we only need to do it once. Now that I think of it we should probably have a separate test exclusively for smsbot

app/verifier/tests/test_api_views.py
```diff
@@ -115,17 +130,27 @@ def test_proper_targetsmart_request(requests_mock, mocker, settings):
...
-    patched_sms_send = mocker.patch("verifier.api_views.send_welcome_sms")
-    settings.SMS_OPTIN_REMINDER_DELAY = 150
-    settings.SMS_POST_SIGNUP_ALERT = True
+def test_smsbot_trigger(requests_mock, smsbot_patch):
+    client = APIClient()
+    targetsmart_call = requests_mock.register_uri(
+        "GET", TARGETSMART_ENDPOINT, json={"result": [], "too_many": False},
+    )
+
+    client.post(LOOKUP_API_ENDPOINT, VALID_LOOKUP)
```
nickcatal, line 145: @liewegas what are your thoughts about just splitting it out into a separate test?

PR #142, opened 2020-05-19
official: add /v1/official/address/ endpoint to find correct region

liewegas:
- switch to postgis in dev environment
- uses PointField for address location
- for most states matches based on county, city, town, etc.
- tweaks for a few states (NY, AL) where things aren't quite by county
- for WI, queries a state arcgis API to get the exact municipality name, since the municipality and mailing address may not match
- when necessary (mostly rural new england), fall back to providing a list of the closest LEOs by distance, closest first.
- sync with USVF nightly
- NOTE: the first USVF sync will take 2-3 hours to populate the new column

TODO
- [x] add structured address query ability? (`street`, `city`, etc. vs a single `q` string)
- [x] get some manual testing with more WI, MI, and NYC addresses
- [ ] incorporate test cases that hit live APIs... i'm not sure where the ``test_address.py`` in the last commit should go
- [ ] move copious comments documenting state corner cases somewhere else?


app/official/api_views.py
```diff
@@ -46,3 +34,201 @@ class RegionDetailViewSet(mixins.RetrieveModelMixin, viewsets.GenericViewSet):
...
+        # narrow search to ~70 miles in each direction.  except Alaska, which has weird
+        # (but few) regions, and is so sparse that a distance filter is dangerous
+        if state_code != "AK":
+            queryset = queryset.filter(lat__gte=(lat - 1)).filter(lat__lte=(lat + 1))
+            queryset = queryset.filter(lng__gte=(lng - 1)).filter(lng__lte=(lng + 1))
+
+        for office_addr in queryset:
+            if not office_addr.lat or not office_addr.lng:
+                continue
+            dist = distance(lat, lng, office_addr.lat, office_addr.lng)
```
jlev, line 198: how fast is this? I'll bet we can decrease load by offloading to PostGIS
we can add geodjango Point fields to the Address model for the latlon to make this easier

app/official/models.py
```diff
@@ -50,6 +50,9 @@ class Address(USVFModel):
     phone = PhoneNumberField(null=True)
     fax = PhoneNumberField(null=True)
 
+    lat = models.FloatField(null=True)
+    lng = models.FloatField(null=True)
```
jlev, line 54: these will perform better as a combined Point
`from django.contrib.gis.geos import Point`

jlev: make sure to add `geography=True` in the definition to ensure the right spherical calculation that will return meters instead of degrees
https://docs.djangoproject.com/en/3.0/ref/contrib/gis/model-api/#geography-type

app/official/api_views.py
jlev: example using geodjango: https://stackoverflow.com/a/52251277

liewegas:
Updated with lots of changes.  I also have a bunch of testcases (~20) that are real addresses and expected results but since they hit live external APIs i'm not sure how to incorporate that into the repo...

app/official/api_views.py
```diff
@@ -46,3 +44,293 @@ class RegionDetailViewSet(mixins.RetrieveModelMixin, viewsets.GenericViewSet):
...
+            # quantize distance to an int so that we can use it as a dict key
+            dist = int(1000000.0 * dist)
+            if dist not in by_distance:
+                by_distance[dist] = [office_addr.office.external_id]
+            else:
+                by_distance[dist].append(office_addr.office.external_id)
+
+        closest_offices = []
+        for dist in sorted(by_distance.keys())[0:max_by_distance]:
+            closest_offices.extend(by_distance[dist])
```
jlev, line 298: can we do this sorting in the geo query above?
we'd be able to avoid converting to a dict until we have just the three closest

app/turnout/settings/core.py
```diff
@@ -37,7 +37,7 @@
 
 DATABASES = {
     "default": env.dj_db_url(
-        "DATABASE_URL", default="pgsql://postgres:turnout@postgres:5432/turnout"
+        "DATABASE_URL", default="postgis://postgres:turnout@postgres:5432/turnout"
```
jlev, line 40: let's double check if this is necessary. not sure I've had to do it with other postgis projects
does it have any side effects?

liewegas: until I made this change ``make migrate`` would fail with an error when trying to add the location PointField column

app/official/api_views.py
liewegas: updated!

app/turnout/settings/core.py
jlev: never mind, looks like it's required
https://docs.djangoproject.com/en/3.0/ref/contrib/gis/db-api/

app/common/geocode.py
```diff
@@ -0,0 +1,189 @@
...
+def geocode(**kwargs):
+    args = {}
+    for k in ["street", "city", "state", "q", "fields"]:
+        if k in kwargs:
+            args[k] = kwargs[k]
+    if "zipcode" in kwargs:
+        args["postal_code"] = kwargs["zipcode"]
+    url = f"{API_ENDPOINT}?{urlencode({**args, 'api_key': settings.GEOCODIO_API_KEY})}"
+    with statsd.timed("turnout.common.geocode.geocode", sample_rate=0.2):
+        r = requests.get(url)
```
benweissmann, line 24: Do you have a sense of how reliable this endpoint is? Should we use `requests`'s retry mechanism to give this a couple tries?

app/common/geocode.py
```diff
@@ -0,0 +1,189 @@
...
+    if r.status_code != 200:
+        return None
```
benweissmann, line 26: This should get logged (both via `logger` and via `sentry_sdk`)

app/common/geocode.py
```diff
@@ -0,0 +1,189 @@
...
+        "returnGeometry": "false",
+        "outSR": "4326",
+        "f": "json",
+        "geometry": f"{location.x},{location.y}",
+    }
+    url = f"{API_ENDPOINT}?{urlencode(args)}"
+    with statsd.timed("turnout.common.geocode.wi_latlng_to_mcd", sample_rate=0.2):
+        r = requests.get(url)
+    if r.status_code != 200:
+        return None
```
benweissmann, line 48: Ditto on logging/retrying... I don't have a lot of confidence in state websites to handle spikes in traffic during GOTV -- is there something we can fall back to here if the state website falls over?

app/official/api_views.py
```diff
@@ -22,22 +37,6 @@ def get_queryset(self):
         if municipality:
             queryset = queryset.filter(municipality=municipality)
 
-        absentee = self.request.query_params.get("absentee", False)
```
benweissmann, line 25: Can you say more about why you removed this? Will this cause issues until we switch over to using the new endpoint on the frontend?

test_address.py
```diff
@@ -0,0 +1,80 @@
+import pytest
+
+from urllib.parse import urlencode
+import requests
+
+API_URL = "http://localhost:9001/v1/official/address/"
+
+# NOTE: these tests requires a functioning local stack, including a
```
benweissmann, line 8: Instead of hitting a real DB and real external APIs, I'd suggest using mock data -- we use [django model bakery](https://pypi.org/project/model-bakery/) to populate our fixtures during tests, and [VCR](https://vcrpy.readthedocs.io/en/latest/usage.html) looks like a good way to capture all the HTTP request/responses and then cache them in the repo.

benweissmann: I think having this version that hits the real stuff is good too though! Especially if we're trying to track down a bug that we think might be caused by upstream API changes. Maybe we can have two versions: one that runs as part of the normal test suite and uses mock DB entries / API responses, and then this version that hits the real things that we mark with [skip](https://docs.pytest.org/en/latest/skipping.html#skip) but can run manually.

benweissmann:
One other suggestion: it would be great to capture everything you've learned about the various edge cases and data sources you had to pull (maybe a google doc?). In particular, we should document what things are likely to change, and what we'll need to keep an eye on to keep up to date, especially for things that might change before November, or in 2021 redistricting.

app/common/geocode.py
liewegas: can you point to an example of the sentry_sdk use in turnout?

app/common/geocode.py
liewegas: In this case it will fall back to a distance search (if there is no exact match).

app/official/api_views.py
liewegas: @nickcatal mentioned that it didn't end up being valuable and could be removed.

app/common/geocode.py
benweissmann: Yup, here's one I just added on my branch: https://github.com/vote/turnout/blob/feature/esign-faxing/app/absentee/tasks.py#L76

app/official/api_views.py
benweissmann: 👍

liewegas:
@benweissmann Okay, pushed all my updates.  This uses geocodio for the easy states and TS for WI, MI, and most of New England where things are extra irritating.  I tested with a disabled TS API (simulated that endpoint is down) and only a handful of tests fail (returning multiple nearby results instead of just one).

Added logging and sentry etc.

I started to mock a request but it *also* has to mock all of the database content and TBH I don't think it is worth it--I'd have to do several cases to cover most of the code paths.  And running the end-to-end tests is easy to do locally anyway.

test_address.py
```diff
@@ -0,0 +1,99 @@
...
+        ("46640 Healy St", "Dodgeville", "MI", "49921", [{'name': 'Portage Township, Houghton County', 'external_id': 436123}]),
+
+        # ny
+        ("20 W 34th St", "New York", "NY", "10001", [{'name': 'New York City: Manhattan', 'external_id': 433320}]),
+        ("1050 Clove Rd", "Staten Island", "NY", "10301", [{'name': 'New York City: Staten Island', 'external_id': 433318}]),
+        ("5800 20th Ave", "Brooklyn", "NY", "11204", [{'name': 'New York City: Brooklyn', 'external_id': 433317}]),
+        ("1825 Eastchester Rd", "The Bronx", "NY", "10461", [{'name': 'New York City: The Bronx', 'external_id': 433316}]),
+        ("71-01 Parsons Blvd", "Fresh Meadows", "NY", "11365", [{'name': 'New York City: Queens', 'external_id': 433319}]),
+        ("535 Morgan Ave", "Brooklyn", "NY", "11222", [{'name': 'New York City: Brooklyn', 'external_id': 433317}]),
+        ("11 Hazen St", "East Elmhurst", "NY", "11370", [{'name': 'New York City: The Bronx', 'external_id': 433316}]),
```
benweissmann, line 94: Could you add a couple test cases for addresses that don't have good matches and we have to fall back to returning N nearest results?

liewegas: with the addition of the TS endpoint i haven't been able to find any such addresses.  the only time it would happen now is if TS is down or times out and it falls back to geocod.io in the annoying states like VT or NH

PR #145, opened 2020-05-19
Add db view that allows easy access to action statistics

nickcatal:
We store data about our user's action status inside a large "events" table that we're constantly appending to. This code creates a postgres view that aggregates that data on a per-action basis, so we can get general details like "has this person finished the action" and "how many times did this person download the PDF that they were sent"

I also added an unmanaged django model with a OneToOne relationship to Action, so that you can easily access this data using the django ORM.


benweissmann:
Looks good it me! I ran an `EXPLAIN ANALYZE` on the underlying SQL on our prod DB and it looks pretty reasonable.

PR #148, opened 2020-05-20
Increase number of gunicorn workers

nickcatal:
Instead of 1 sync worker running, we should run num cores * 2 + 1. That should let us scale better.

We'll want to look into gevent very soon since verifier is IO-bound, but that's a bigger issue.


app/turnout/gunicorn.conf.py
```diff
@@ -1,7 +1,11 @@
+import multiprocessing
+
 timeout = 75
 keepalive = 75
 accesslog = "-"
 errorlog = "-"
+num_proc = multiprocessing.cpu_count()
```
benweissmann, line 7: Do we know how this interacts with Fargate? Like does this return the CPU count based on the allocated CPUs, or the underlying hardware? And if it's the former, couldn't this give us a fractional result?

nickcatal: I tested this. 4 vCPUs triggers 9 workers.

PR #156, opened 2020-05-24
Restrict users to only access their own dashboards

nickcatal:
Previously all users could access all client dashboards. With this change users can only access the specific partners they are authorized to access.


app/accounts/manage_profile_views.py
```diff
@@ -8,16 +9,32 @@
...
+    def dispatch(self, request, *args, **kwargs):
+        if not request.user.multi_client_admin:
+            raise Http404
+        return super().dispatch(request, *args, **kwargs)
+
     def get_redirect_url(self, *args, **kwargs):
         client = get_object_or_404(
             Client, default_slug__slug=self.kwargs["partner_slug"]
         )
+        if client not in self.request.user.allowed_clients:
```
benweissmann, line 36: Does this compile into some sort of efficient SQL? Or does this end up loading all allowed_clients into memory?

app/accounts/manage_profile_views.py
```diff
@@ -8,16 +9,32 @@
...
     context_object_name = "partners"
     template_name = "multi_tenant/manage/change_partner_list.html"
 
+    def dispatch(self, request, *args, **kwargs):
+        if not request.user.multi_client_admin:
+            raise Http404
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_queryset(self):
+        return self.request.user.allowed_clients.select_related(
```
benweissmann, line 21: We might want a follow-up ticket to figure out some sort of pagination so this doesn't need to return every client for superadmins.

Maybe we could have paginated views of clients like we have for other models, that lets us view/edit clients, and we could also put a button there to switch to that client?

app/accounts/manage_profile_views.py
nickcatal: It loads all `allowed_clients` into memory. It's a cached property though, so theoretically that query should only run once regardless of how many times we mention the property throughout our codebase or in templates.

app/accounts/manage_profile_views.py
nickcatal: 👍

app/accounts/manage_profile_views.py
benweissmann: Instead of `allowed_clients()`, could we just have an `is_allowed_client()` function that runs an `EXISTS` query? Not super pressing right now but maybe something to put in a follow-up ticket

benweissmann, line 36: Should this be a `POST`? Would be a little harder to wire up on the frontend, but it seems like having a `GET` write to the database is likely to causing caching/retry issues. On the frontend, we should be able to wrap the whole set of buttons in a `<form>` and just use `name`/`value` properties on each button so the browser will tell us which button was clicked

nickcatal: That would be optimal, but you'd make `request.user.allowed_clients` not accessible inside of django templates. I don't think this is a HUGE issue since it will apply exclusively to superadmins who have rights on every org. Also, we don't store a huge amount of data inside the `Client` model.

benweissmann: Yeah, I think it's something to come back to once we have a bunch of clients and need to rework that screen to use pagination

nickcatal: It indeed should be a `POST`

app/multi_tenant/forms.py
```diff
@@ -0,0 +1,7 @@
+from django import forms
+
+from .models import Client
+
+
+class ChangePartnerManageForm(forms.Form):
+    partner = forms.ModelChoiceField(queryset=Client.objects.none())
```
benweissmann, line 7: What does the `.none()` do here?

app/multi_tenant/manage_views.py
```diff
@@ -1,7 +1,33 @@
...
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_form(self):
+        form = super().get_form()
+        form.fields["partner"].queryset = self.request.user.allowed_clients
+        return form
+
+    def form_valid(self, form):
+        new_client = form.cleaned_data["partner"]
+        self.request.user.active_client = new_client
```
benweissmann, line 31: Do we need a validation here to make sure the new_client is in the allowed_clients? Or does the queryset in `get_form` handle that? Either way, would be good to have a comment here with how the authorization check works.

app/multi_tenant/forms.py
nickcatal: You can't have a `ModelChoiceField` without a QuerySet. But we set the queryset explicitly inside of `get_form`

app/multi_tenant/manage_views.py
nickcatal: The queryset in `get_form` handles that.

nickcatal: Added comment about this.

app/multi_tenant/forms.py
nickcatal: Added comment about this.

PR #158, opened 2020-05-24
Add LEO contact overrides to the admin panel

benweissmann:
Add an admin UI for the LEO contact overrides.

The done janky thing is that you have to put in the region ID manually when creating a new record -- the django default renders a dropdown of all regions, which is really slow. You can get the region ID pretty easily from the LEO contact lookup tool (it's in the URL) so this should be OK for now since it's just us that will be using this.

There is currently no access control on these pages -- but it looks like that's true of the election info pages too so I assume we're going to be adding access control to all the non-partner-specific pages before partner launch.


nickcatal:
Do we want to add this to the management interface or the django admin? And how do we want to control access to this?

app/absentee/models.py
```diff
@@ -79,4 +81,4 @@ class LeoContactOverride(TimestampModel):
     fax = PhoneNumberField(null=True)
 
     class Meta:
-        ordering = ["-created_at"]
+        ordering = ["region__state__code", "region__name"]
```
nickcatal, line 84: Did this not create a migration?

benweissmann:
I was thinking we'd add it to the management interface and ACL it however we ACL the state information editor -- my understanding was that the django admin is really just for the dev team, and the management interface is for stuff that we want other folks on the team to work with -- this feels like something that we could let program or research work with directly

app/absentee/models.py
benweissmann: Oh, it looks like it does, just committed that. Apparently it doesn't actually touch the DB, it just tracks it for if you want to run python code in a historical context: https://docs.djangoproject.com/en/3.0/ref/migration-operations/#altermodeloptions

nickcatal: Yeah, I usually just throw the change in a previous migration so that django's tracking doesn't cause an unnecessary migration.

app/manage/urls.py
```diff
@@ -8,6 +8,7 @@
     path("logout/", views.LogoutView.as_view(), name="logout"),
     path("profile/", include("accounts.manage_profile_urls")),
     path("election/", include("election.manage_urls")),
+    path("absentee/leo_contact_overrides/", include("absentee.manage_override_urls")),
```
nickcatal, line 11: @benweissmann can you make this `absentee_overrides/`? then there won't be a 404 at `absentee/`

PR #159, opened 2020-05-26
Rename "partner" to "subscriber"

nickcatal:
To reflect the change in nomenclature, change "partner" to "subscriber" throughout the app

This does not change the structure of the database, but it does change the parameters expected.

Of particular note, anywhere that we have `?partner=` in the URL we now need `?subscriber=`, and `sms_opt_in_partner` is now `sms_opt_in_subscriber`

**Note: To run this without doing weird stuff with the database you'll want to fake the migrations, you can do this via to `./manage.py migrate --fake`**


jlev:
Might be worth adding
```
        verbose_name = "Subscriber"
        verbose_name_plural = "Subscribers"
``` to Client model Meta
just for consistency in the admin

PR #161, opened 2020-05-27
manage: a few basic stats on subscriber page

liewegas:


app/multi_tenant/models.py
```diff
@@ -41,6 +42,10 @@ def full_email_address(self) -> str:
         clean_name = self.name.replace('"', "'")
         return f'"{clean_name}" <{self.email}>'
 
+    @property
+    def stats(self):
+        return cache.get(f"client.stats/{self.uuid}")
```
nickcatal, line 47: What happens if there is a cache miss (perhaps the client has been created in the past 5 minutes, or something funky happened in redis)

app/reporting/tasks.py
```diff
@@ -20,3 +30,78 @@ def process_report(report_pk: str):
...
+    n = 0
+    for client in Client.objects.all():
+        calc_subscriber_stats.apply_async((client.uuid,), countdown=n)
+        n += 2
+
+
+@shared_task
+@statsd.timed("turnout.reporting.calc_subscriber_stats")
+def calc_subscriber_stats(uuid):
+    logger.info("calc_subscriber_stats %s" % uuid)
```
nickcatal, line 48: Random request, can you make this (taken from https://github.com/vote/turnout/blob/master/app/mailer/tasks.py#L29-L36 )

```
    extra = {
        "subscriber_uuid": uuid,
    }
    logger.info(
        "Calculated Subscriber Stats %(subscriber_uuid)s", extra, extra=extra
    )
```

The `extra=` is going to let us do fancy filtering inside of datadog

app/reporting/tasks.py
```diff
@@ -20,3 +30,78 @@ def process_report(report_pk: str):
...
+    r = {}
+
+    # some totals
+    r["register"] = (
+        Registration.objects.filter(subscriber_id__exact=uuid)
```
nickcatal, line 53: In theory you shouldn't need to do `__exact` here. `__exact` is implied with the =.

For future reference, since it's Postgres and capitalization matters, if you ever want to do case insensitive matching, `__iexact` is great.

app/reporting/tasks.py
```diff
@@ -20,3 +30,78 @@ def process_report(report_pk: str):
...
     trigger_notification(report)
+
+
+@shared_task
+@statsd.timed("turnout.reporting.calc_all_subscriber_stats")
+def calc_all_subscriber_stats():
+    logger.info("Recalculating all subscriber stats")
+    n = 0
+    for client in Client.objects.all():
+        calc_subscriber_stats.apply_async((client.uuid,), countdown=n)
```
nickcatal, line 41: Perfect!

Can you set an expires on this as well? Maybe base it on the length of time between each run which we can set in settings?

I worry about our queue getting full and a ton of these firing at once. Probably better to just have it miss running the analytics in any one cycle and fall back to old analytics than have our database hit with a huge number of analytics runs when the queue starts to clear up.

app/manage/templates/management/manage_home.html
```diff
@@ -6,9 +6,34 @@
...
+	  </tr>
+	  {% endfor %}
+	  <tfoot>
+	    <td>Total</td>
+	    <td>{{stats.register}}</td>
+	    <td>{{stats.verify}}</td>
+	    <td>{{stats.absentee}}</td>
+	  </tfoot>
+	</tbody>
+      </table>
```
nickcatal, line 35: Can you throw a "last updated" timestamp in here? Maybe include a line "Analytics are delayed up to 5 minutes".

You can make any date inside a template timezone aware by doing `{{ stats.last_updated|timezone:"America/Los_Angeles" }}`

We'll probably add timezone aware-ness soon, but for now we just use LA.

app/reporting/tasks.py
```diff
@@ -20,3 +30,78 @@ def process_report(report_pk: str):
...
+        .filter(action__details__finished__exact=True)
+        .values("state_id")
+        .annotate(count=Count("state_id"))
+        .order_by("state_id")
+    ):
+        if i["state_id"] not in r["by_state"]:
+            r["by_state"][i["state_id"]] = {}
+        r["by_state"][i["state_id"]]["absentee"] = i["count"]
+
+    cache.set(f"client.stats/{uuid}", r)
```
nickcatal, line 107: Do you think it makes sense to set a timeout here, incase a client is deleted? Maybe like 24 hours or something?

app/reporting/tasks.py
```diff
@@ -20,3 +30,78 @@ def process_report(report_pk: str):
...
+@shared_task
+@statsd.timed("turnout.reporting.calc_subscriber_stats")
+def calc_subscriber_stats(uuid):
+    logger.info("calc_subscriber_stats %s" % uuid)
+    r = {}
+
+    # some totals
+    r["register"] = (
+        Registration.objects.filter(subscriber_id__exact=uuid)
+        .select_related("action", "actiondetails")
```
nickcatal, line 54: To do a select related of a table 1 table away, use `.select_related("action", "action__actiondetails")`

nickcatal: Sorry, `select_related("action", "action__details")`

app/multi_tenant/models.py
liewegas: add ``or {}`` and verified that the caller can cope

app/reporting/tasks.py
liewegas: added ``CALC_STATS_INTERVAL`` to core.py.. not sure if there is a better name for that

PR #162, opened 2020-05-27
User Management

nickcatal:
This adds

1) A "Profile management" page where a user can change their name
2) A password reset page where a user can reset their password
3) A page which lists all the users associated with any one subscriber
4) A page which allows users to be sent invites
5) Pages where users and invites can be removed from individual clients
6) A "roadblock" page that users who have valid logins but no valid or current subscriber sees. If a user is associated with multiple subscribers, and they're removed from the subscriber they're actively connected to, they'll be shown the roadblock and given an option which subscriber they wish to switch to.
7) Transactional emails that are sent to end-users
8) Timezone support on a per-user basis (does not apply to reports)


app/accounts/migrations/0006_user_timezone.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('accounts', '0005_active_client_nullable'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='user',
+            name='timezone',
+            field=timezone_field.fields.TimeZoneField(choices=[('US/Alaska', 'US/Alaska'), ('US/Arizona', 'US/Arizona'), ('US/Central', 'US/Central'), ('US/Eastern', 'US/Eastern'), ('US/Hawaii', 'US/Hawaii'), ('US/Mountain', 'US/Mountain'), ('US/Pacific', 'US/Pacific'), ('UTC', 'UTC')], default='US/Eastern', null=True),
```
benweissmann, line 17: We should probably make the default Pacific because previously we were showing everything in Pacific time

app/absentee/templates/absentee/manage/ballot_request_list.html
```diff
@@ -31,7 +31,7 @@
                                 {{ ballot_request.state.pk }}
                             </td>
                             <td>
-                                {{ ballot_request.created_at|timezone:"America/Los_Angeles" }}
+                                {{ ballot_request.created_at|timezone:user.timezone }}
```
benweissmann, line 34: Can we change this date format to include the timezone, or put the user's timezone in the column header? Especially if we aren't localizing the exported time, I think it'll be clearer to show that this time is localized and show what timezone it is (especially because I think it would also be reasonable to assume that this is showing the time localized to the person submitting the ballot request rather than to the admin viewing the page)

app/multi_tenant/manage_views.py
```diff
@@ -34,3 +55,216 @@ def form_valid(self, form):
...
+                extra,
+                extra=extra,
+            )
+
+            messages.success(
+                self.request,
+                f"Invite {existing_invite.email} has been added or updated.",
+            )
+
+            return HttpResponseRedirect(
```
benweissmann, line 189: Should we re-sent the transactional invite email here? If the invitation is old the user might have forgotten about it.

app/absentee/templates/absentee/manage/ballot_request_list.html
nickcatal: This is done, also spelled it out in the reports as UTC

app/multi_tenant/manage_views.py
```diff
@@ -34,3 +55,216 @@ def form_valid(self, form):
...
+                "user_uuid": existing_user.pk,
+                "manager_uuid": self.request.user.pk,
+            }
+            logger.info(
+                "User Management: %(manager_uuid)s added %(user_uuid)s to subscriber %(subscriber_uuid)s",
+                extra,
+                extra=extra,
+            )
+
+            return HttpResponseRedirect(
```
benweissmann, line 152: We might want a transactional email here with a link to the page that lets you switch between subscribers, and like a short blurb that you've been added to a new subscriber -- I think a lot of users might not have used a system where you use one account and switch between subscribers

benweissmann:
Couple notes:

- Is there a way to reset your password without logging in?
- I was able to accept an invite and log in without setting up 2fa -- did you remove 2fa in preparation for doing SMS 2fa?
- Navigating to any page other than the root page or the login page while not logged in throws a 500 error (`'AnonymousUser' object has no attribute 'active_client'`). Similarly, only the homepage shows the roadblock, not any of the other pages.

nickcatal:
1) not yet, will work on this soon
2) That's likely due to your local environment
3) That's a bug, will work on it separately

PR #166, opened 2020-05-29
Integrate fax gateway into Turnout

benweissmann:
I need to do some more manual testing before merging this (mostly making sure that there's no regressions for the existing LEO email and self-print functionality) but this is ready for review.

- New `fax` application that sends faxes via the fax gateway and handles callbacks
- Associated settings/env vars for faxing
- Rework `generateform.py` to support generating fax-ready documents, and switch to faxing for GA esign rather than email
- Add Debra's fax cover sheet
- Add transactional emails and Celery jobs for esign faxing


app/absentee/tests/test_metadata.py
```diff
@@ -148,3 +148,15 @@ def test_slugs_match_pdf(state):
             assert (
                 "On" in field["FieldStateOption"]
             ), f"Expected PDF field {field['FieldName']} to use \"On\" as its checked value"
+
+
+# Make sure page counts are correct
+@pytest.mark.parametrize("state", STATES_WITH_METADATA)
```
nickcatal, line 154: Does it make sense to parallelize our tests at this point?

app/common/enums.py
```diff
@@ -120,7 +122,9 @@ class Labels:
         START = "Started Form"
         FINISH_SELF_PRINT = "Finished via Print and Mail"
         FINISH_EXTERNAL = "Finished with External Tool"
-        FINISH_LEO = "Finish via direct LEO submission"
+        FINISH_LEO = "Finish via direct LEO email submission"
+        FINISH_LEO_FAX_PENDING = "Finish via direct LEO fax submission (fax pending)"
+        FINISH_LEO_FAX_SENT = "Finish via direct LEO fax submission (fax sent)"
```
nickcatal, line 127: We need to make sure we update the `ActionDetail` postgres view to reflect these new finish events, otherwise reports and exports won't be correct.

app/fax/api_views.py
```diff
@@ -0,0 +1,72 @@
...
+            return Response(
+                {"error": "Token is incorrect"}, status=status.HTTP_403_FORBIDDEN
+            )
+
+        if fax.status_timestamp and (
+            serializer.validated_data["timestamp"] < fax.status_timestamp
+        ):
+            return Response(
+                {"warning": "Outdated timestamp; ignoring"}, status=status.HTTP_200_OK
+            )
```
nickcatal, line 68: Probably want to pipe incorrect-token and outdated timestamps events to statsd and logs.

app/fax/migrations/0001_initial.py
```diff
@@ -0,0 +1,35 @@
...
+import django.db.models.deletion
+import django_smalluuid.models
+
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+        ('storage', '0005_rename_partner_subscriber'),
```
nickcatal, line 15: Can you have this rely on the storage migration `0001_initial` instead of `0005_rename_partner_subscriber`

I like giving the django migration runner as much flexibility as possible when choosing the order in which to apply migrations. And the best way to do that is to have migrations depend on the earliest migration possible.

app/fax/send_fax.py
```diff
@@ -0,0 +1,85 @@
...
+    fax = Fax(
+        storage_item=storage_item,
+        to=to,
+        status=FaxStatus.PENDING,
+        on_complete_task=on_complete_task,
+        on_complete_task_arg=on_complete_task_arg,
+    )
+    fax.save()
+
+    callback_url = f"{settings.FAX_GATEWAY_CALLBACK_URL}?token={str(fax.token)}"
```
nickcatal, line 45: Is the `str()` needed here?

app/turnout/settings/core.py
```diff
@@ -510,11 +511,30 @@
...
+
+
+#### FAX CONFIGURATION
+
+FAX_DISABLE = env.bool("FAX_DISABLE", default=True)
+
+FAX_OVERRIDE_DEST = env.str("FAX_OVERRIDE_DEST", default=None)
+
+FAX_GATEWAY_CALLBACK_URL = env.str(
+    "FAX_GATEWAY_CALLBACK_URL", default=f"{PRIMARY_ORIGIN}/fax/gateway_callback/"
```
nickcatal, line 535: Just a flag that currently `PRIMARY_ORIGIN` will 301 to the API origin, and will result in the callback going through Cloudflare's WAF.

We don't want this traffic going through the public internet if we can help it.

app/absentee/leo_fax.py
```diff
@@ -0,0 +1,100 @@
...
+
+    msg.content_subtype = "html"
+    msg.send()
+
+
+def send_fax_and_fax_submitted_email(ballot_request: BallotRequest) -> None:
+    leo_contact = get_absentee_contact_info(ballot_request.region.external_id)
+
+    if not leo_contact.fax:
+        raise NoAbsenteeRequestFaxAddress()
```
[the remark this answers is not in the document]
benweissmann: Tracking in follow up: https://app.clubhouse.io/turnout2020/story/1016/how-do-we-want-to-ensure-that-we-have-the-email-fax-for-states-where-we-re-doing-esign

app/absentee/tests/test_metadata.py
benweissmann: Tracking in follow-up: https://app.clubhouse.io/turnout2020/story/1038/parallelize-travis-builds

PR #170, opened 2020-06-01
Allow restrictions on editing of election information

nickcatal:
Users who are not explicitly set to be election admins in the django administration should not have access to that functionality in the management interface.


app/election/mixins_manage_views.py
```diff
@@ -0,0 +1,8 @@
+from django.http import Http404
+
+
+class ElectionAdminView:
+    def dispatch(self, request, *args, **kwargs):
+        if not request.user.can_manage_election_information:
+            raise Http404
```
benweissmann, line 7: Should this be 403?

nickcatal: I'm thinking we'll just want to hide the existence of these pages.

PR #174, opened 2020-06-03
Infrastructure improvements (gevent, postgis, datadog)

nickcatal:
- Add support for gevent workers, including properly patching postgres and adding proper pooling

- Installing postgis, which is needed for upcoming projects

- Updated datadog's tracer to the latest version, which offers a lot more visibility into app performance.


app/turnout/settings/core.py
```diff
@@ -40,6 +41,9 @@
         "DATABASE_URL", default="pgsql://postgres:turnout@postgres:5432/turnout"
```
liewegas, line 41: s/pgsql/postgis/

app/turnout/settings/core.py
```diff
@@ -40,6 +41,9 @@
...
     )
 }
+DATABASES["default"]["ENGINE"] = "django_db_geventpool.backends.postgis"
```
liewegas, line 44: ...can be used instead of this

app/turnout/settings/core.py
```diff
@@ -78,7 +82,6 @@
     "rest_framework",
```
liewegas, line 82: need to add to DJANGO_APPS:
https://github.com/vote/turnout/pull/142/commits/ef06f478eeef12cb5ea2f83ea9349b86559575fb#diff-ecc3487e6c1cffd39606c2a61272c195R70

app/turnout/settings/core.py
nickcatal: We need to use geventpool for our move to gevent. Can't use the built in postgis library

PR #180, opened 2020-06-04
Improvements to absentee API

benweissmann:
- Support `?incomplete=true` for `PATCH` requests
- Support a new `?match_region=true` parameter for absentee requests that will auto-select the correct region based on Sage's work
- If we can't match to a single region, try to filter the list of regions we present to the user (only if `match_region` is passed; if you don't pass `match_region` everything should work exactly as before for backwards-compatibility)
- Create a new `esign_method` field on the BallotRequest model, and populate it based on the region and USVF data availability. Return it to the frontend on `incomplete` calls.

This should support the new esign workflow. We can now `PATCH` as many times as we want, passing `incomplete=true` until we're ready for final submission. After the initial POST, we'll return either:

```json
{ "uuid": "...", "action_id": "...", "region": 123456, "esign_method": "leo_email" }
```

if we were able to auto-select the region, or 

```json
{ "uuid": "...", "action_id": "...", "region": null, "esign_method": null, "regions": [...] }
```

if we need to ask the user to pick a region. Then, when you `PATCH` to update it with the region, you'll get back the filled-in `esign_method` in response to the `PATCH`.

The frontend should use `esign_method` to determine whether it can show the esign flow. If it's `leo_email` or `leo_fax`, and the user is on mobile, we show the esign flow. Otherwise, we show the self-print flow.

cc @camdenlee


app/official/api_views.py
```diff
@@ -344,7 +344,7 @@ def ts_address_to_region(street, city, state, zipcode):
     except KeyError:
         extra = {"url": url, "response": r.json()}
         logger.error(
-            "Malformed result from TS district %(url)s, response %(response)s",
+            "Malformed result from TS district %{url}s, response %{response}s",
```
nickcatal, line 347: We fixed this in another PR, but these should not be `%{}` but instead `%()`

app/official/api_views.py
```diff
@@ -465,5 +463,21 @@ def address_regions(request):
...
+def address_regions(request):
+    regions, was_geocode_error = get_regions_for_address(
+        street=request.query_params.get("street", None),
+        city=request.query_params.get("city", None),
+        state=request.query_params.get("state", None),
+        zipcode=request.query_params.get("zipcode", None),
+    )
+
+    if was_geocode_error:
+        return Response("unable to geocode address", status=status.HTTP_400_BAD_REQUEST)
```
nickcatal, line 480: do we pass this to statsd and log it anywhere? I think we do in `get_regions_for_address()`

app/absentee/api_views.py
```diff
@@ -1,41 +1,134 @@
...
+            regions, _ = get_regions_for_address(
+                street=f"{serializer.validated_data['address1']} {serializer.validated_data.get('address2') or ''}",
+                city=serializer.validated_data["city"],
+                state=serializer.validated_data["state"],
+                zipcode=serializer.validated_data["zipcode"],
+            )
+
+            if not regions or len(regions) == 0:
+                # No region matching data -- user has to pick
+                pass
```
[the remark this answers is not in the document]
benweissmann: Yup, we do that below -- we pass here because we *also* need return all the regions if the user didn't pass `match_region=true`, so we just leave the state as-is and handle it below.

app/official/api_views.py
benweissmann: Oh weird; I don't remember editing this. Maybe a weird merge thing; will revert

app/official/api_views.py
benweissmann: Added

app/absentee/tests/test_api_views.py
```diff
@@ -27,14 +34,57 @@
...
+
+
+@pytest.fixture
+def mock_get_regions_for_address(mocker):
+    return mocker.patch("absentee.api_views.get_regions_for_address")
+
+
+@pytest.fixture
+def process_ballotrequest_submission_task(mocker):
+    return mocker.patch("absentee.api_views.process_ballotrequest_submission.delay")
```
nickcatal, line 68: I haven't been including the `.delay` in my patches, then having my code in the test be `process_ballotrequest_submission_task.delay.assert_called_with()`.

This allows us to test calls to both `delay` and `apply_async` using 1 patch, since there are times when we use both ways to invoke a task, depending if we need celery to do some sort of magic or not.

app/action/mixin_apiview.py
```diff
@@ -31,6 +31,9 @@ def create(self, request, *args, **kwargs):
 
         return self.create_or_update_response(request, action_object)
 
+    def after_create_or_update(self, action_object):
+        pass
```
nickcatal, line 35: I haven't been doing this (creating a stub method.) Is there some reason you're doing it here? mypy passing, style preference, etc?

It's a good idea and I can start doing it, just wondering.

app/absentee/tests/test_api_views.py
benweissmann: I think that doesn't work here because we're passing `process_ballotrequest_submission_task` itself around -- it's not called in this file but rather in the parent class.

app/action/mixin_apiview.py
benweissmann: The registration views don't override this (this isn't a DRF method; it's something that I added that's called below in between when we apply the model updates, and when we queue the follow-up job) so we need a stub here because not all child classes need to implement this.

app/absentee/tests/test_api_views.py
nickcatal: Okay!

PR #186, opened 2020-06-05
feature: official/tasks: check for valid contact info after USVF sync

liewegas:
- usvf sync runs the check after the sync.
- check_vbm_states() is a separate task so it can be run independently
- check_state_contacts takes the vbm_submission_type, so that a state can
be check for a mode that is not currently set (to see whether the change is
safe to make).


liewegas:
This can be run manually with something like
``
make shell
echo -e 'from common import enums\nimport official.tasks\nprint(official.tasks.check_state_contacts("WI", enums.SubmissionType.LEO_FAX))' | ./manage.py shell
``

app/official/tasks.py
```diff
@@ -1,8 +1,101 @@
...
 
 @shared_task
 def sync_usvotefoundation():
     from .usvf_sync import sync
 
     sync()
+    check_vbm_states()
+
+
+def check_state_contacts(state_id: str, mode: enums.SubmissionType) -> str:
```
nickcatal, line 26: Can we move this out of `tasks.py`?

app/official/tasks.py
```diff
@@ -1,8 +1,101 @@
...
+    return None
+
+
+@shared_task
+def check_vbm_states():
+    for state in State.objects.all():
+        if state.vbm_submission_type != "self_print":
+            message = check_state_contacts(state.code, state.vbm_submission_type)
+            if message:
+                logger.warning(message)
```
nickcatal, line 93: I believe this is going to pipe the error to Sentry. Do we want this behavior?

app/official/tasks.py
```diff
@@ -1,8 +1,101 @@
...
+                if settings.SLACK_DATA_ERROR_ENABLED:
+                    try:
+                        r = requests.post(
+                            settings.SLACK_DATA_ERROR_WEBHOOK, json={"text": message}
+                        )
```
nickcatal, line 98: Can we pipe monitoring of this to statsd?

app/official/tasks.py
```diff
@@ -1,8 +1,101 @@
...
+
+logger = logging.getLogger("official")
 
 
 @shared_task
 def sync_usvotefoundation():
     from .usvf_sync import sync
 
     sync()
+    check_vbm_states()
```
nickcatal, line 23: Do we want to run this with `.delay()` or run it inline in the task?

app/official/tasks.py
liewegas: where should this go?  i'm assuming not usvf_sync.py.  new file?  (validate.py?)

app/official/tasks.py
liewegas: I think so?  If this triggers it's the live db that has the problem and it needs quick attention (vs `check_state_contacts()` which can be run manually for any state + mode)

app/official/tasks.py
liewegas: do you mean something like
```
statsd.increment("turnout.official.missing_leo_contact_info")
```
?

app/official/tasks.py
liewegas: Inline seems simpler.  The check is fast (esp compared to the sync).

app/official/tasks.py
nickcatal: `usvf_validate.py`?

app/official/tasks.py
nickcatal: Then is there a reason to leave it in `tasks.py`? Might make sense to throw in `usvf_validate.py`

app/official/tasks.py
nickcatal: Makes sense to me. let's see if it gets annoying or not

app/official/tasks.py
liewegas: i could merge it all into usvf.py..

nickcatal: Not a bad idea at all

PR #187, opened 2020-06-05
Add docs for the region geocoding endpoint; make parameter names match

benweissmann:
- Add docs for `/official/address/`
- Rename `street` to `address1` to match other endpoints


app/official/api_views.py
```diff
@@ -473,8 +475,9 @@ def get_regions_for_address(street, city, state, zipcode):
 
 @api_view(["GET"])
 def address_regions(request):
+    print(request.query_params)
```
liewegas, line 478: stray line

PR #189, opened 2020-06-08
Allow actions to be searched by email and name

nickcatal:
There is a need to be able to quickly search actions to find an individual record. Thankfully django has built-in support for postgres search and our 3 actions (Verify, Register, Absentee) all share 1 common ListView, so this is relatively easy.

The styling isn't the fanciest, but it works.

[ch1109]


app/action/mixin_manage_views.py
```diff
@@ -1,5 +1,17 @@
...
 class ActionListViewMixin:
     paginate_by = settings.MANAGEMENT_ACTION_PAGINATION_SIZE
+
+    def get_queryset(self):
+        queryset = super().get_queryset()
+
+        search_query = self.request.GET.get("q")
+        if search_query:
+            queryset = queryset.annotate(
+                search=SearchVector("first_name", "last_name", "email")
```
benweissmann, line 14: Couple thoughts:

- I'm not sure the postgres tsvectors are quite what we want here -- tsvectors are great for searching blocks of text and handling stemming, but they're not very useful for names/emails. I'm not sure this is going to give us more useful search that a simpler `first_name || ' ' || last_name || ' ' || email ILIKE '%' || escaped_query || '%'`. GIven the django integration, tsvector might be simpler? But also `ILIKE` will give us partial-text matches

- In either case, we'll want an index to speed this up -- a GiST index on either the `to_tsvector` expression if we're using tsvectors, or on a `pg_trgm` expression if we're using `ILIKE` (and then we'd want to limit search terms to 3 characters or more)

PR #190, opened 2020-06-08
bugfix: official/usvf: workaround buggy USVF API pagination

liewegas:
Make multiple passes over the USVF API with different page sizes to ensure
we capture all records.  In practices, 2 passes with 100 and 73 is
sufficient, but make a third pass as an insurance policy.

I don't have high confidence that USVF will fix this problem on their end quickly.  :/


nickcatal:
Can you add some sort of django setting to let us turn on/off this sync, or change the number of cycles? Having staging, prod, dev, and anyone who leaves docker-compose running overnight hitting the USVF database all at the same time is... just not nice.

liewegas:
> Can you add some sort of django setting to let us turn on/off this sync, or change the number of cycles? Having staging, prod, dev, and anyone who leaves docker-compose running overnight hitting the USVF database all at the same time is... just not nice.

Updated.  The prod/staging/dev ``.env`` files need to be updated with
```
USVF_SYNC=True
USVF_SYNC_HOUR=something
```

I didn't change the number of passes since I think we always want to do all 3 to get a correct result...

PR #191, opened 2020-06-08
Update transactional emails

benweissmann:
- Move the action buttons to the top of the email and improve the copy a bit
- Add horizontal rules to break up the emails
- Left-align everything in the emails
- Add donation asks for first-party emails
- Slight adjustment of alert styles in the admin tool to make them more legible


app/multi_tenant/migrations/0013_client_is_first_party.py
```diff
@@ -0,0 +1,31 @@
...
+UPDATE multi_tenant_client
+SET is_first_party = TRUE
+WHERE url ILIKE 'https://www.voteamerica.com%';
+"""
+
+UNSET_FIRST_PARTIES = """
+UPDATE multi_tenant_client
+SET is_first_party = NULL
+WHERE url ILIKE 'https://www.voteamerica.com%';
+"""
```
nickcatal, line 15: Instead of relying on SQL I'd use the Django ORM similar to https://github.com/vote/turnout/blob/master/app/election/migrations/0002_load_states.py here

The more you can rely on the ORM, the easier it is to migrate to new database engines down the line if you're unlucky enough to have to do that.

PR #193, opened 2020-06-09
Improve accuracy of /v1/official/address

liewegas:
now only ~10 bad matches from 40k addresses in prod, all of them bad input, except one case of a weird AFB address


liewegas:
Added support for importing the data sets into postgis... basically just need to run scripts/import_region_geom.sh with the proper postgres creds and then add this to ``.env``:
```
GEOCODE_USE_GIS_TABLE_MA=True
GEOCODE_USE_GIS_TABLE_WI=True
```

```

File: app/common/management/commands/importgisdata.py [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+
+        # WI
+        try:
+            cursor.execute("drop table official_region_geom_wi")
+            cursor.commit()
+        except:
+            pass
+        os.system(
+            f"cd {td.name} && wget https://opendata.arcgis.com/datasets/47009ab057d8401eb9025c4fc024a12a_0.zip && unzip 47009ab057d8401eb9025c4fc024a12a_0.zip && shp2pgsql -s 4326 CTV_Jan_2020.shp official_region_geom_wi | {psql}"
+        )
```
nickcatal, line 27: Can you pull the data first, then only drop the table if the get was successful?

app/common/geocode.py
```diff
@@ -75,38 +84,127 @@ def wi_location_to_mcd(location):
...
+    if settings.GEOCODE_USE_GIS_TABLE_MA:
+
+        class PrecinctMA(models.Model):
+            # note: these are only the fields we care about
+            gid = models.IntegerField(primary_key=True)
+            town = models.TextField()
+            geom = MultiPolygonField(geography=True)
+
+            class Meta:
+                db_table = "official_region_geom_ma"
```
nickcatal, line 189: I'd actually add an unmanaged model that maps to the columns of the table you're importing, similar to https://github.com/vote/turnout/blob/master/app/action/models.py#L17-L33

Then we can access that table however we wish.

With an unmanaged model django's migration system will never touch the database, django will just assume a table or view exists that matches the schema of the model and make it your responsibility to ensure that all the fields in the model exist in the table/view.

nickcatal:
> I'm not sure the import should go to a celery beat, though--if the data set does change it will have a new source URL and filename, so the import really is a one-time thing and data set changes will require human intervention (_also_ true if we're hitting the live API, which specifies the dated data set to query).
> 
> And I think dropping the live API makes sense... if the data set changes we need to updated regardless--either what API we hit or to grab the new data. @nickcatal what do you think?

100% with the idea to drop the live API if possible.

If we can move this task into celery, at least for cases where it's not local dev, we can run it on the worker, which gives us data in datadog and sentry regarding performance/errors/etc that the entire team can review. Also bandwidth inside the VPC is much faster (and in theory more reliable) than through the VPN.

If it's a beat then we can't forget about doing it on staging/prod/dev. But if that's not possible that's not possible.

nickcatal:
Actually, if this is a one-time thing doing it locally using a django management command is fine.

app/common/management/commands/importgisdata.py
liewegas: fixed

app/common/geocode.py
liewegas: fixed, now in ``official/models.py``

PR #200, opened 2020-06-13
Add full-text search indices

benweissmann:
- Add this package, which handles setting up the indices and triggers for full-text indexing: https://github.com/damoti/django-tsvector-field. The simple installation method led to this: https://github.com/damoti/django-tsvector-field/issues/3 so I used the other installation method (which the docs say is more reliable anyway).

- Add search fields to ballot requests, registrations, and verifications

- Use the new search field in the admin backend

I verified (with `EXPLAIN`) that the new GIN indices are used for these searches.


app/register/models.py
```diff
@@ -80,6 +81,15 @@ class Registration(
...
     referring_tool = TurnoutEnumField(enums.ToolName, null=True, blank=True)
 
+    search = tsvector_field.SearchVectorField(
+        [
+            tsvector_field.WeightedColumn("email", "A"),
+            tsvector_field.WeightedColumn("last_name", "A"),
+            tsvector_field.WeightedColumn("first_name", "B"),
+        ],
+        "english",
+    )
```
nickcatal, line 91: Since this identical code exists in 3 models, we probably want to throw it in an abstract model

app/turnout/settings/core.py
```diff
@@ -41,7 +41,7 @@
         "DATABASE_URL", default="postgis://postgres:turnout@postgres:5432/turnout"
     )
 }
-DATABASES["default"]["ENGINE"] = "django_db_geventpool.backends.postgis"
+DATABASES["default"]["ENGINE"] = "turnout_db"
```
nickcatal, line 44: Do we want to put this in `/app/turnout_db/base.py` or `/app/common/db/base.py`?

I've been throwing random stuff like this (analytics, apm, etc) in the `common` directory

app/common/utils/models.py
```diff
@@ -27,3 +29,21 @@ class TrackingModel(models.Model):
 
     class Meta(object):
         abstract = True
+
+
+class SearchableEmailNameModel(models.Model):
+    first_name = models.TextField(null=True)
+    last_name = models.TextField(null=True)
+    email = models.EmailField(null=True)
```
nickcatal, line 37: Super random Q that may not matter: Does it make sense to move these columns off of the action models themselves?

app/absentee/models.py
```diff
@@ -3,23 +3,31 @@
...
 
 
 class BallotRequest(
-    ActionModel, SubscriberModel, TrackingModel, UUIDModel, TimestampModel
+    ActionModel,
+    SubscriberModel,
+    TrackingModel,
+    UUIDModel,
+    TimestampModel,
+    SearchableEmailNameModel,
```
nickcatal, line 26: Super random request, can you put `SearchableEmailNameModel` before `TimestampModel`. Every other model in the app ends with `UUIDModel, TimestampModel`

app/common/utils/models.py
benweissmann: So i initially tried to leave these fields on the model, but was getting `common.SearchableEmailName.search: (postgres.E100) No textual columns available in this model for search vector indexing.` and some related errors about the `SearchableEmailName` model referencing fields that weren't in the model. I thought the only way to fix this was to move the fields to the searchable model, but it looks like marking the model abstract may also fix this. Update (maybe) coming shortly.

app/absentee/models.py
benweissmann: Done -- I ran `make lint`, which i think alphabetizes this.

app/common/utils/models.py
benweissmann: Fixed!

PR #203, opened 2020-06-15
sync to actionnetwork

liewegas:
- Please confirm that the Link model makes sense.  It allows a different external id linkage for each (registration or ballotrequest or lookup).  I didn't tie this to Event, but certainly could.
- The current version will populate the `identifiers` list on actionnetwork for a person (keyed by email) with links to *each* lookup, registration, and absentee event on our side.  So my test user for example has
```
        "identifiers": [
          "action_network:9f2be294-b86c-4c9a-a050-10057110dc19",
          "voteamerica_lookup:iOBqSYJETUCoB-0TzRNoJg",
          "voteamerica_lookup:dXw1LJw2SwW1f3lVsXzdsw",
          "voteamerica_lookup:14OR5NVmRl2rRd2N2MyUjg",
          "voteamerica_lookup:KUv5o6TrSo-wGQ0AaRjvVw",
          "voteamerica_lookup:nSt3q0wpSui6sS3yDBYmXA",
          "voteamerica_lookup:lSL7gJIzRKGgc3C80QdTJg"
        ],
```
- There is currently no attempt to dedup on our side--it's last writer wins.  So if `user@foo.com` does a lookup with one address and then registers with another address, the last one of them will be the address on ActionNetwork.  That probably won't happen much, but perhaps we want to prefer the registration over lookup...

TODO
- [x] confirm data model makes sense
- [ ] add phone #'s (once actionnetwork API is updated); respect sms_opt_in flag 
- [x] clean up error handling, log exceptions to sentry, etc.
- [x] trigger event for registration
- [x] trigger event for ballotrequest
- [ ] add migrations to pr


app/integration/models.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Link(TimestampModel):
+    lookup = models.ForeignKey("verifier.Lookup", null=True, on_delete=models.SET_NULL)
+    registration = models.ForeignKey(
+        "register.Registration", null=True, on_delete=models.SET_NULL
+    )
+    ballotrequest = models.ForeignKey(
+        "absentee.BallotRequest", null=True, on_delete=models.SET_NULL
+    )
```
liewegas, line 13: We might want to CASCADE instead?  Although most likely the parent records will never get deleted..

app/multi_tenant/models.py
```diff
@@ -73,3 +73,15 @@ class Association(UUIDModel, TimestampModel):
...
     client = models.ForeignKey("multi_tenant.Client", on_delete=models.CASCADE)
     user = models.ForeignKey("accounts.Invite", on_delete=models.CASCADE)
+
+
+class SubscriberIntegrationProperty(TimestampModel):
+    subscriber = models.ForeignKey(
+        Client, on_delete=models.CASCADE, db_column="partner_id"
+    )
+    external_tool = models.TextField(null=True)
+    name = models.TextField(null=True)
```
nickcatal, line 83: Perhaps we use enumfields here?

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,189 @@
...
+        if ids.startswith("action_network"):
+            return ids.split(":", 1)[1]
+    return None
+
+
+@statsd.timed("turnout.integration.sync_lookup")
+def sync_lookup(lookup):
+    _sync_lookup(lookup, None)
+    if lookup.subscriber_id:
+        _sync_lookup(lookup, lookup.subscriber_id)
```
nickcatal, line 38: Every action (lookup, register, absentee) has a subscriber, even 1st party ones.

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,189 @@
...
+
+ADD_ENDPOINT = "https://actionnetwork.org/api/v2/people/"
+
+logger = logging.getLogger("integration")
+
+
+def post_person(info, api_key):
+    response = requests.post(
+        ADD_ENDPOINT, json=info, headers={"OSDI-API-Token": api_key},
+    )
```
nickcatal, line 23: Can you do something like https://github.com/vote/turnout/blob/master/app/mailer/tasks.py#L21-L22 where it wraps the request in a new service in datadog

```python
from common.apm import tracer
with tracer.trace("an.person.add", service="actionnetwork"):
    requests.post()
```

app/integration/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Link(TimestampModel):
+    lookup = models.ForeignKey("verifier.Lookup", null=True, on_delete=models.CASCADE)
+    registration = models.ForeignKey(
+        "register.Registration", null=True, on_delete=models.CASCADE
+    )
+    ballotrequest = models.ForeignKey(
+        "absentee.BallotRequest", null=True, on_delete=models.CASCADE
+    )
```
nickcatal, line 14: You can just foreignkey to `action.Action`, which all 3 of these link to

app/multi_tenant/models.py
```diff
@@ -73,3 +73,15 @@ class Association(UUIDModel, TimestampModel):
 class InviteAssociation(UUIDModel, TimestampModel):
     client = models.ForeignKey("multi_tenant.Client", on_delete=models.CASCADE)
     user = models.ForeignKey("accounts.Invite", on_delete=models.CASCADE)
+
+
+class SubscriberIntegrationProperty(TimestampModel):
```
nickcatal, line 78: can you also make this a `UUIDModel`

Otherwise there is an auto-incrementing primary key added to the database, and guessable primary keys is less than optimal when using a framework that likes to use primary keys in URLs

app/verifier/api_views.py
```diff
@@ -127,6 +128,8 @@ def create(self, request, *args, **kwargs):
                 args=(str(instance.phone),), countdown=settings.SMS_OPTIN_REMINDER_DELAY
             )
 
+        sync_lookup_to_actionnetwork.delay(instance.uuid)
```
nickcatal, line 131: Can we make this more reproducible? Nearly identical code has to exist in the `api_views.py` files for absentee, register, and lookup, and all of these API views are actions that somehow link to the `action/` folder. And they all have basically the same fields. Just a thought, maybe there is no way to make this reproducible.

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,189 @@
...
+        api_key = None
+        for key in SubscriberIntegrationProperty.objects.filter(
+            subscriber_id=subscriber_id,
+            external_tool="actionnetwork",
+            name="api_key"):
+            api_key = key.value
+        if not api_key:
+            logger.info("No actionnetwork api_key for subscriber %s" % subscriber_id)
+            return
+    id_ = post_person(
```
nickcatal, line 54: Random AF minor request: Can you use something more descriptive here than `id_` to describe what this variable is?

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,189 @@
...
+
+@statsd.timed("turnout.integration.sync_lookup")
+def sync_lookup(lookup):
+    _sync_lookup(lookup, None)
+    if lookup.subscriber_id:
+        _sync_lookup(lookup, lookup.subscriber_id)
+
+
+def _sync_lookup(lookup, subscriber_id):
+    logger.info("Sync Lookup %s, subscriber %s" % (lookup, subscriber_id))
```
nickcatal, line 42: Can you use f-strings or the built-in logger string generation?

app/verifier/api_views.py
liewegas: i collapsed all the code in ``actionnetwork.py`` since it's all keys on action_id now, and the other field names match regardless of the type, but for the front end of the process we still need to pull the correct type of record.

nickcatal: This makes sense for lookups, but for BallotRequests and Registrations I'd follow the same way the other tasks are done

app/absentee/api_views.py
```diff
@@ -111,6 +112,7 @@ def create(self, request, *args, **kwargs):
 
     def after_create_or_update(self, ballot_request: BallotRequest):
         populate_esign_method(ballot_request)
+        sync_ballotrequest_to_actionnetwork.delay(ballot_request.uuid)
```
nickcatal, line 115: I believe this will figure on both incomplete submissions and complete submissions.

I'd riff off what we do in IncompleteActionViewset

We set a local `task` property on the Absentee and Register views https://github.com/vote/turnout/blob/master/app/absentee/api_views.py#L45

Then inside the IncompleteActionViewset mixin we fire `self.task.delay()` ( https://github.com/vote/turnout/blob/master/app/action/mixin_apiview.py#L74-L75 )

Register and Absentee are VERY similar, so the more they can share code the easier it is going to be for us to keep complete test coverage and add integrations that support both

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,154 @@
...
+            logger.info(f"No actionnetwork api_key for subscriber {subscriber_id}")
+            return None
+    return settings.ACTIONNETWORK_KEY
+
+
+@statsd.timed("turnout.integration.actionnetwork.sync_item")
+def sync_item(item):
+    if settings.ACTIONNETWORK_SYNC:
+        _sync_item(item, None)
+        _sync_item(item, item.subscriber_id)
```
nickcatal, line 82: Most subscribers won't have an action network integration, so we probably want to do a test here to see if the subscriber is using action network or not before going down this path.

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,154 @@
...
+
+def get_api_key(subscriber_id):
+    if subscriber_id:
+        api_key = None
+        for key in SubscriberIntegrationProperty.objects.filter(
+            subscriber_id=subscriber_id, external_tool="actionnetwork", name="api_key"
+        ):
+            return key.value
+        if not api_key:
+            logger.info(f"No actionnetwork api_key for subscriber {subscriber_id}")
```
nickcatal, line 73: Wouldn't this end up flooding the logs, since most subscribers won't have active integrations?

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,154 @@
...
+        if ids.startswith("action_network"):
+            return ids.split(":", 1)[1]
+    return None
+
+
+def get_api_key(subscriber_id):
+    if subscriber_id:
+        api_key = None
+        for key in SubscriberIntegrationProperty.objects.filter(
+            subscriber_id=subscriber_id, external_tool="actionnetwork", name="api_key"
```
nickcatal, line 69: Since we're using an EnumField now this can be

`external_tool=ExternalToolType.ACTIONNETWORK`

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,154 @@
...
+        ):
+            return key.value
+        if not api_key:
+            logger.info(f"No actionnetwork api_key for subscriber {subscriber_id}")
+            return None
+    return settings.ACTIONNETWORK_KEY
+
+
+@statsd.timed("turnout.integration.actionnetwork.sync_item")
```
nickcatal, line 78: Instead of using statsd.timed can you do `@tracer.wrap()` like https://github.com/vote/turnout/blob/master/app/action/serializers.py#L32

We'll get the same details and a more granular flame graph in datadog APM

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+
+
+def sync_all_items(cls):
+    for item in (
+        cls.objects.annotate(
+            no_sync=~Exists(
+                Link.objects.filter(
+                    subscriber_id=None,
+                    external_tool="actionnetwork",
```
nickcatal, line 123: We can use `ExternalToolType.ACTIONNETWORK` here

app/multi_tenant/models.py
```diff
@@ -73,3 +73,15 @@ class Association(UUIDModel, TimestampModel):
 class InviteAssociation(UUIDModel, TimestampModel):
     client = models.ForeignKey("multi_tenant.Client", on_delete=models.CASCADE)
     user = models.ForeignKey("accounts.Invite", on_delete=models.CASCADE)
+
+
+class SubscriberIntegrationProperty(UUIDModel, TimestampModel):
+    subscriber = models.ForeignKey(
+        Client, on_delete=models.CASCADE, db_column="partner_id"
```
nickcatal, line 80: I don't know that we need to explicitly set `db_column` here.

app/multi_tenant/models.py
```diff
@@ -73,3 +73,15 @@ class Association(UUIDModel, TimestampModel):
...
+    )
+    external_tool = models.TextField(null=True)
```
nickcatal, line 82: This should be an EnumField with the `ExternalToolType` enum

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,154 @@
...
+                )
+            )
+        )
+        .filter(no_sync=True)
+        .order_by("created_at")
+    ):
+        _sync_item(item, item.subscriber_id)
+
+
+@statsd.timed("turnout.integration.actionnetwork.sync")
```
nickcatal, line 149: Can you do `@tracer.wrap()` here

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,154 @@
...
+        .order_by("created_at")
+    ):
+        _sync_item(item, None)
+
+    for item in (
+        cls.objects.annotate(
+            no_sync=~Exists(
+                Link.objects.filter(
+                    subscriber_id=OuterRef("subscriber"),
+                    external_tool="actionnetwork",
```
nickcatal, line 138: We can use `ExternalToolType.ACTIONNETWORK` here

app/integration/actionnetwork.py
```diff
@@ -0,0 +1,154 @@
...
+
+@statsd.timed("turnout.integration.actionnetwork.sync_item")
+def sync_item(item):
+    if settings.ACTIONNETWORK_SYNC:
+        _sync_item(item, None)
+        _sync_item(item, item.subscriber_id)
+
+
+def _sync_item(item, subscriber_id):
+    logger.info(f"Sync {item}, subscriber {subscriber_id}")
```
nickcatal, line 86: Can you pass `{subscriber_id}` as an `extra` item to the logger, like we use elsewhere?

app/verifier/tests/test_api_views.py
```diff
@@ -189,6 +194,7 @@ def test_proper_alloy_request(requests_mock):
 
 @pytest.mark.django_db
 def test_lookup_object_created(requests_mock, mocker, smsbot_patch):
+    settings.ACTIONNETWORK_SYNC = False
```
nickcatal, line 197: Oh, one last thing here, we can just disable ACTIONNETWORK_SYNC once using the autoused pytest fixture at the top of the test file

https://github.com/vote/turnout/blob/master/app/verifier/tests/test_api_views.py#L78-L82

app/integration/actionnetwork.py
liewegas: We don't know until we look up the api key in ``get_api_key``, which happens right at the top of ``_sync_item``.

app/integration/actionnetwork.py
liewegas: yep, removed

PR #208, opened 2020-06-17
Add DD profiler

benweissmann:
I ran this locally and it runs OK but doesn't send profiles -- maybe because it's not run with `ddtrace-run`? I tried running with `ddtrace-run` but got other errors. I think we should try deploying this to dev and see if it works.


nickcatal:
@benweissmann can you split this out into 2PR, one to update the ECS task definition with the API key, and another to test the profiler? you can then do `make localtodev` to push your local version to dev to test things out

PR #212, opened 2020-06-18
/v1/official/address: return no result is the state doesn't match up

liewegas:
If the user enters inconsistent information (e.g., a state and zip that
don't match up), and the address we match is in the wrong state, then
return *no* result to the user.  The frontend can fall back to providing
a list of all regions in the state.


app/official/api_views.py
```diff
@@ -139,6 +139,9 @@ def geocode_to_regions(street, city, state, zipcode):
 
     # assume the first address match is the correct (only) state
     state_code = addrs[0]["address_components"]["state"]
+    if state_code != state:
+        # if the address match gets the state wrong, return *no* result
```
benweissmann, line 143: We should log these so we can search through the logs and get some better data on how common this is and what the error usually is.

PR #213, opened 2020-06-18
Add additional APM tracing

nickcatal:
In order to get more visibility in the datadog flame graph we need to wrap key methods we care about in `@tracer.wrap()`. I included anything that may involve processing time, such as hitting external services or rendering templates. Monitoring email compile time is going to be especially useful if we ever decide to get more complex with our transactional emails or throw more fields in.

The other advantage here is we'll be able to link individual queries in the flame graph to components of the request, instead of the entire view.


benweissmann:
It looks like most of these are paired with a call to `statsd.timed` -- can we make a utility function that does both? It seems like if something it worth timing it's also worth tracing and vice-versa.

nickcatal:
@benweissmann it turns out that the `statsd.timed()` calls are completely redundant and unnecessary. Everything that is wrapped is also available as a metric inside of datadog.

PR #222, opened 2020-06-24
Add support for custom OVR link integrations with states; add configs for WA and CO

benweissmann:
- Add hooks for returning a customize OVR link for the user from a registration `POST`. We store this in the DB so it can be reused (for WA and CO it's a cheap idempotent function to generate the URL, but for CA we're going to need to hit the state API and get a custom token).

- Add support for WA and CO. CA coming soon and NV hopefully coming at some point!


ecs/template/include/secrets.libsonnet
```diff
@@ -153,5 +153,17 @@
...
+      if env == 'prod' then [
+        {
+          valueFrom: 'arn:aws:ssm:us-west-2:719108811834:parameter/turnout.' + env + '.register_co_vrd_id',
+          name: 'REGISTER_CO_VRD_ID',
+        },
+        {
+          valueFrom: 'arn:aws:ssm:us-west-2:719108811834:parameter/turnout.' + env + '.register_wa_vrd_id',
+          name: 'REGISTER_WA_VRD_ID',
+        },
+      ] else []
```
nickcatal, line 167: This is going to make `remote_run.sh` really difficult to handle. Is there no way to have this on all of the versions? maybe just have the ID 00000 or something on staging/dev?

PR #229, opened 2020-06-27
bugfix: actionnetwork: skip items with no action

liewegas:


app/integration/actionnetwork.py
```diff
@@ -85,6 +85,8 @@ def sync_item(item):
 
 
 def _sync_item(item, subscriber_id):
+    if not item.action:
+        return
```
nickcatal, line 89: Actually, can you just filter out where `action` is null from the `sync_all` below?

liewegas: updated!

PR #230, opened 2020-06-29
feature: integration: monitor external_tool_* sites

liewegas:
Push these to datadog synthetics so that we gather uptime data.

- [ ] add on/off flag to ecs env

Next steps (post- this PR):
- wait for twitter (group) dev account to be approved
- calculate update over given period based on ``.../results`` request (unfortunately datadog does not provide summary stats for you--it's all calculated client-side)
- sort out when the twitter bot should tweet (when a site goes down?  when it comes back up, so that we have a new downtime percentage?  after its been down for X long?  etc.)

### DIFF
--- a/app/integration/datadog_synthetics.py
+++ b/app/integration/datadog_synthetics.py
@@ -1,0 +1,112 @@
+import copy
+import logging
+
+import datadog
+import requests
+from django.conf import settings
+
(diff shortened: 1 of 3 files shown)


liewegas:
see https://app.datadoghq.com/synthetics/list

PR #235, opened 2020-07-01
chore: no exceptions for failed address lookups

liewegas:
- Don't throw exceptions
- warning, not error


app/official/api_views.py
```diff
@@ -572,21 +572,11 @@ def get_regions_for_address(street, city, state, zipcode):
         regions = geocode_to_regions(street, city, state, zipcode)
 
     if not regions:
-        logger.error(f"No match for ({street}, {city}, {state} {zipcode})")
-        sentry_sdk.capture_exception(
-            AddressRegionsAPIError(
-                f"No match for ({street}, {city}, {state} {zipcode})"
-            )
-        )
```
nickcatal, line 580: Do we even need the `AddressRegionsAPIError` exception anymore?

PR #239, opened 2020-07-01
bugfix: fix actionnetwork more

liewegas:


liewegas:
ugh. i think adding mypy would be a worthwhile investment of effort...

PR #241, opened 2020-07-01
bugfix: make logger log more than just the message

liewegas:
We want other key details like the logger name, module, file/func/line,
level, etc.


liewegas:
Other keys I left out:
``
            'asctime',
            'created',
            'levelno',
            'msecs',
            'pathname',
            'process',
            'processName',
            'relativeCreated',
            'thread',
            'threadName'
``

PR #242, opened 2020-07-01
use PA OVR API

liewegas:


app/register/models.py
```diff
@@ -90,6 +91,9 @@ class Registration(
 
     referring_tool = TurnoutEnumField(enums.ToolName, null=True, blank=True)
 
+    state_fields = JSONField(null=True, blank=True)
```
benweissmann, line 94: This needs a migration, no?

Also maybe `state_api_result` instead of `state_api_details`?

app/register/api_views.py
```diff
@@ -1,22 +1,148 @@
...
+logger = logging.getLogger("register")
+
 
 class RegistrationViewSet(IncompleteActionViewSet):
     model = Registration
     serializer_class = RegistrationSerializer
     queryset = Registration.objects.filter(status=TurnoutActionStatus.INCOMPLETE)
-    task = process_registration_submission
+
+    def process_pa_registration(self, registration, state_id_number, is_18_or_over):
```
benweissmann, line 28: Would be good to have some unit tests of this function -- you can just mock out `geocode` and `ovrlib` (with something like `mocker.patch('register.api_views.geocode')` -- [example](https://github.com/vote/turnout/blob/057db58d31faed08b5b0f77c7c977df881e57024/app/mailer/tests/test_tasks.py#L23))

app/register/api_views.py
```diff
@@ -1,22 +1,148 @@
...
+            # FIXME: are these implied by is_18_or_over ???
+            united_states_citizen=True,
+            declaration=True,
+        )
+        logger.info(req)
+        try:
+            res = session.register(req)
+            logger.info(res)
+            registration.state_api_details = {
+                "application_id": res.application_id,
```
benweissmann, line 113: Maybe `"status": "success"` here to make it easier to query based on status?

app/register/api_views.py
```diff
@@ -1,22 +1,148 @@
...
+            api_key=settings.PA_OVR_KEY, staging=settings.PA_OVR_STAGING
+        )
+        req = ovrlib.pa.PAOVRRequest(
+            **r,
+            party=str(registration.party),
+            mailing_address=registration.previous_address1,
+            previous_address=registration.previous_address1,
+            federal_voter=state_fields.get("federal_voter", False),
+            eighteen_on_election_day=is_18_or_over,
+            # FIXME: are these implied by is_18_or_over ???
```
benweissmann, line 104: I don't think so -- we should require the frontend to pass these in in `state_fields` (the frontend should validate that they're always True, but I think explicitly passing them minimizes the chance that a frontend bug lets you submit without checking these boxes)

app/register/api_views.py
```diff
@@ -30,6 +156,7 @@ def create_or_update_response(self, request, action_object):
             "uuid": action_object.uuid,
             "action_id": action_object.action.pk,
             "custom_ovr_link": action_object.custom_ovr_link,
+            "state_api_details": action_object.state_api_details,
```
benweissmann, line 159: I'm a little worried about passing through `state_api_details`, because it has a full exception trace for unexpected errors, and I don't really trust the PA API to not put anything sensitive in the response. Can we return something like `state_api_status: "success" | "failure", state_api_error: null | "invalid_dl" | "invalid_signature" | "invalid_ssn" | "unknown_error"`?

app/register/api_views.py
```diff
@@ -1,22 +1,148 @@
...
+        logger.info(req)
+        try:
+            res = session.register(req)
+            logger.info(res)
+            registration.state_api_details = {
+                "application_id": res.application_id,
+                "signature_source": res.signature_source,
+            }
+            registration.status = TurnoutActionStatus.PENDING
+        except ovrlib.exceptions.InvalidDLError as e:
```
benweissmann, line 117: Is there an error code for an invalid SSN?

app/register/api_views.py
```diff
@@ -1,22 +1,148 @@
...
+        logger.info(addrs)
+        if not addrs:
+            registration.state_api_details = {
+                "status": "address_invalid",
+            }
+            return
+        r["county"] = addrs[0]["address_components"]["county"]
+        if r["county"].endswith(" County"):
+            r["county"] = r["county"][0 : -len(" County")]
+        logger.info(r)
```
benweissmann, line 91: Do we want these `logger.info`s? Or should they be `logger.debug`?

app/register/api_views.py
liewegas: only for badly formatted input (not 4 digits).  the above exception's error text is actually "Please supply either a PA driver's license or PennDOT ID card number, the last four digits of your SSN, or click the check box"

liewegas:
Okay, I think the main remaining item is what the SSN behavior should be.  In the PA staging environment, if I submit with a DL number, it will return an error if the PennDOT lookup fails.  If I do a SSN, it just says "registration accepted with no signature", which IIUC means you then need to mail some paper form within 2 weeks. 

I think we should just not support the SSN case at all--I'm not sure what purpose it serves.  Either they provide a valid DL number and we look it up successfully, or we do the esign thing.  I think that means the wireframe should change a bit to drop the SSN option entirely?

benweissmann:
Sounds good to me!

app/register/api_views.py
```diff
@@ -1,23 +1,163 @@
...
+        )
+        req = ovrlib.pa.PAOVRRequest(
+            **r,
+            party=str(registration.party),
+            mailing_address=registration.previous_address1,
+            previous_address=registration.previous_address1,
+            federal_voter=state_fields.get("federal_voter", False),
+            eighteen_on_election_day=is_18_or_over,
+            united_states_citizen=registration.us_citizen,
+            declaration=declaration,
```
benweissmann, line 107: The UI mocks have a checkbox for "I was previously registered with a different party" -- it doesn't look like we're using that; can we get rid of it?

benweissmann:
@liewegas I've updated the mocks to remove SSN and also clarify the flow / error-handling behavior. Can you take a look over the revised mocks? https://balsamiq.cloud/suqa7b0/piwvpef/rD0E7

app/register/api_views.py
```diff
@@ -1,23 +1,163 @@
...
+                "previous_first_name",
+                "previous_middle_name",
+                "previous_last_name",
+                "previous_city",
+                "previous_state",
+                "previous_zipcode",
+                "mailing_city",
+                "mailing_state",
+                "mailing_zipcode",
+                "gender",
```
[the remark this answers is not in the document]
liewegas: it's optional; the front-end can leave it off

app/register/api_views.py
```diff
@@ -1,23 +1,163 @@
...
+            return
+        r["county"] = addrs[0]["address_components"]["county"]
+        if r["county"].endswith(" County"):
+            r["county"] = r["county"][0 : -len(" County")]
+
+        # attempt registration
+        session = ovrlib.pa.PAOVRSession(
+            api_key=settings.PA_OVR_KEY, staging=settings.PA_OVR_STAGING
+        )
+        req = ovrlib.pa.PAOVRRequest(
```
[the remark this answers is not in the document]
liewegas: It's supported by ovrlib.  I'll add it here.

app/register/api_views.py
```diff
@@ -1,23 +1,163 @@
...
+            registration.state_api_result = {
+                "status": "failure",
+                "error": "unknown_error",
+                "exception": str(e),
+            }
+
+    def complete(
+        self, serializer, registration, state_id_number, is_18_or_over, declaration
+    ):
+        if registration.state_id == "PA" and registration.state_fields:
```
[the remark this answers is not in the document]
liewegas: yep!

app/register/api_views.py
liewegas: Sigh.. yeah.  It's required according to the PA spreadsheet.  Adding!

app/register/api_views.py
```diff
@@ -1,23 +1,163 @@
...
+            r["ssn4"] = state_id_number
+        else:
+            registration.state_api_result = {
+                "status": "failure",
+                "error": "no_dl_or_signature",
+            }
+            return
+
+        # geocode the county
+        addrs = geocode(
```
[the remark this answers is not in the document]
liewegas: fixed

liewegas:
> @liewegas I've updated the mocks to remove SSN and also clarify the flow / error-handling behavior. Can you take a look over the revised mocks? https://balsamiq.cloud/suqa7b0/piwvpef/rD0E7

The only issue I see is that there are checkboxes on the mock for "i have lived at this address ..." and "i am legally qualified to vote", plus the big section and "I agree", but we only check for a single "declaration".  Either we roll those last 2 items into the big section, or the front-end verifies that all of them are checked to make the `declaration` field true?

benweissmann:
Sorry, I should have read more into the SSN stuff earlier -- where are you seeing they they have to mail a form if they use SSN instead of Driver's License? Thinking about it more, I'm worried that the folks we tend to focus on for our programs (younger voters, new voters, and voters of color) may be less likely to have a drivers license than they are to have a social security number (see: http://www.projectvote.org/wp-content/uploads/2015/06/AMERICANS-WITH-PHOTO-ID-Research-Memo-February-2015.pdf) -- so making anyone without a license go through the more-complex esign flow feels like it could disproportionally impact marginalized voters.

benweissmann:
And re: declaration, I think the frontend can just verify that all of those are checked -- I just like having the frontend pass *something* to indicate that the user has definitely ticked all the boxes, so if there's some frontend bug where it fails to handle the declaration section entirely we throw a backend validation error.

app/register/api_views.py
```diff
@@ -1,23 +1,186 @@
...
+                "error": "unknown_error",
+                "exception": str(e),
+            }
+
+    def complete(self, serializer, registration, state_id_number, is_18_or_over):
+        if registration.state_id == "PA" and registration.state_fields:
+            self.process_pa_registration(registration, state_id_number, is_18_or_over)
+            registration.save()
+            if registration.status != TurnoutActionStatus.PENDING:
+                return
```
benweissmann, line 168: We should send a transactional email here with a confirmation + LEO contact info -- something like https://github.com/vote/turnout/blob/master/app/absentee/templates/absentee/email/leo_fax_sent.html (we can skip the inline LEO contact info and just link to https://www.voteamerica.com/local-election-offices/PA/$REGION) -- just something like

```

{% extends "email/base.html" %}
{% load markdownify %}
{% block content %}


{{ recipient.first_name }},<br><br>
Thanks for registering to vote in {{registration.state.name}}.
<br><br>
There's nothing else you need to do. If you have any other questions about your
application, you can reach out directly to your <a href="https://www.voteamerica.com/local-election-offices/{{registration.state.code}}/{{region_id}}/">Local Election Office</a>
  <br>
{% endif %}
<br><br>

<hr><br>

{% if subscriber.is_first_party %}
One last thing: We are a small nonprofit and we don’t take any government funding.
<a href="https://secure.actblue.com/donate/tipjar?email={{registration.email}}&firstname={{registration.first_name}}&lastname={{registration.last_name}}&zip={{registration.zipcode}}&city={{registration.city}}&state={{registration.state.code}}&phone={{registration.phone.as_e164}}&addr1={{registration.address1}}&refcode=absentee_fax_notification" target="_blank">
    Can you chip in $3 to support our work?
</a>
<br><br>
{% endif %}


Happy voting,<br>
{{ subscriber.name }}

{% endblock %}
```

(we should be able to make this generic enough that we can reuse the template/email sending celery task when we implement OVR for other states -- MI docs coming soon!)

liewegas:
> Sorry, I should have read more into the SSN stuff earlier -- where are you seeing they they have to mail a form if they use SSN instead of Driver's License? Thinking about it more, I'm worried that the folks we tend to focus on for our programs (younger voters, new voters, and voters of color) may be less likely to have a drivers license than they are to have a social security number (see: http://www.projectvote.org/wp-content/uploads/2015/06/AMERICANS-WITH-PHOTO-ID-Research-Memo-February-2015.pdf) -- so making anyone without a license go through the more-complex esign flow feels like it could disproportionally impact marginalized voters.

THere is this:

```
Sorry, I should have read more into the SSN stuff earlier -- where are you seeing they they have to mail a form if they use SSN instead of Driver's License? Thinking about it more, I'm worried that the folks we tend to focus on for our programs (younger voters, new voters, and voters of color) may be less likely to have a drivers license than they are to have a social security number (see: http://www.projectvote.org/wp-content/uploads/2015/06/AMERICANS-WITH-PHOTO-ID-Research-Memo-February-2015.pdf) -- so making anyone without a license go through the more-complex esign flow feels like it could disproportionally impact marginalized voters.
```
and this
```
In batch mode, all pertinent application data is submitted within one POST call.
Applications that are submitted in a batch process should allow the applicant to submit both a PennDOT
ID/Driver’s License number and a signature image file. Based on the precedent processing of the PA OVR Web API,
it will first attempt to use the submitted PennDOT ID to retrieve a signature. If the PennDOT ID is validated, the PA
OVR Web API will link the PennDOT signature to the application (and discard the submitted signature file). If the
PennDOT ID cannot be validated, the system will then attempt to process the uploaded signature image file (if
available) to determine if it meets signature quality standards. If neither process can capture the applicant’s
signature, the application will still be submitted into the SURE system, but be placed into a 14-day “hold” status,
pending signature receipt. If a signature is not received after 14 days, the application will be placed into a
“pending” state and the applicant will either receive a signature capture card from their county voter registration
office or missing signature correspondence via the SURE system. Ultimately the applicant must supply a signature
to their county voter registration office to successfully complete his or her registration
```
Not super explicit, but in all of this, I see nothing that suggests that a signature can be retrieved from PennDOT (or anywhere else) with the SSN4.  In fact it's not clear that the SSN4 is used for anything.  I think we have to use the esign flow if there is no DL number.

benweissmann:
Ah, good catch -- I've reached out to Elissa to clarify. Let's go ahead without SSN for right now until we hear back from her.

benweissmann:
One last thing to wrap this up (would be good to get this merged before you leave for the week -- feel free to punt on sending the transactional email for now and we can just file a follow-up ticket in Clubhouse) -- can you write an example request body or a couple examples for DL/SSN/etc? So that we can an example of what the requests should look like coming from the frontend.

liewegas:
> One last thing to wrap this up (would be good to get this merged before you leave for the week -- feel free to punt on sending the transactional email for now and we can just file a follow-up ticket in Clubhouse) -- can you write an example request body or a couple examples for DL/SSN/etc? So that we can an example of what the requests should look like coming from the frontend.

I was going to ask where the API documentation lives... I can put it all there?

benweissmann:
So our public API is here: https://docs.voteamerica.com/api/ (generated from https://github.com/vote/turnout/tree/master/docs), but we don't have great internal API docs... we have some kinda out-of-date docs on Notion here: https://www.notion.so/voteamerica/Register-API-Docs-cca3ecb7a8d74212af3fa865b399ce4b -- so you could add a section to that page or make a new Notion page?

liewegas:
Docs updated--LMK if anything is missing.

Final thought: since we only take DL and not SSN, the ``id_type`` field isn't really needed, right?  Pushed a final patch dropping it, but for the moment it's still in the Notion docs.

PR #243, opened 2020-07-01
bugfix: actionnetwork: no PII in the exception

liewegas:


app/integration/actionnetwork.py
```diff
@@ -53,7 +53,7 @@ def post_person(info, api_key):
         )
         sentry_sdk.capture_exception(
             ActionNetworkError(
-                f"Error posting to {ADD_ENDPOINT}, info {info}, status code {response.status_code}"
+                f"Error posting to {ADD_ENDPOINT}, status code {response.status_code}"
```
nickcatal, line 56: Actually, can you log the info you're removing somewhere?

liewegas: in both cases it's sent to logger.error

nickcatal: Oh great. Carry on then

PR #246, opened 2020-07-02
feature: integration: uptime.com

liewegas:
Public pages.  We don't have much control over what they look like, but uptime.com's public pages seem to be about as good as it gets.  My only complaint is there is a 50 site max, so I leave off DC when we hit that limit.

- https://uptime.com/s/voteamerica-absentee-application
- https://uptime.com/s/voteamerica-absentee-tracker
- https://uptime.com/s/voteamerica-ovr
- https://uptime.com/s/voteamerica-polling-place
- https://uptime.com/s/voteamerica-verify-status

Twitter.  Ignore all the last few tweets, and keep in mind that these sites appear down right now because of the state blocking us (see TODO below)
- https://twitter.com/VA_StateWatch

TODO
- [ ] set ecs vars in AWS
- [ ] convince VT to whitelist uptime.com
- [ ] convince NV to whitelist uptime.com
- [ ] convince AZ to fix the redirect loop on their absentee request page
- [x] confirm ID absentee tracker page no longer exists and then remove it from the database


liewegas:
Sigh..

```
> You agree that Your use of Uptime.com is for the purpose of monitoring
websites either owned and/or controlled by You. You acknowledge that Your use of
Uptime.com for any other purpose, such as to monitor the website of a third
party, is an unauthorized use of Uptime.com and may subject Your account to
termination.
```

PR #257, opened 2020-07-09
Rework reporting views

benweissmann:
- Rewrite the SQL for the action details view to fix the download count, simplify the structure, and add a last-event-at timestamp
- Add three views that correspond to our three reports
- Rework the reporting code to read from those views instead of the models. I used raw SQL for this instead of an unmanaged model so that we don't need to keep the list of column in sync in three different places (we just have the columns in the view SQL, and the CSV column names in the runner)


app/reporting/migrations/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+    action_details.download_count AS total_downloads
+FROM
+    absentee_ballotrequest br
+LEFT JOIN multi_tenant_client subscriber ON br.partner_id = subscriber.uuid
+LEFT JOIN action_actiondetails action_details ON br.action_id = action_details.action_id;
+
+
+CREATE OR REPLACE VIEW reporting_registerreport AS
+SELECT
+    reg.uuid,
```
nickcatal, line 56: My flag here is that this is going to output a uuid, and inside the admin and in our DRF views these actions are going to appear as a smalluuid

I'm not exactly an enormous fan of smalluuids in general and wouldn't strictly be against making switching to normal uuids a task (as long as we do it soon), but we should maintain a standard.

app/action/migrations/__init__.py
```diff
@@ -5,84 +5,30 @@
...
-  a.uuid AS action_id,
-  COALESCE( bool_or(generic_finish.event_exists), bool_or(finish_external.event_exists), bool_or(finish_leo.event_exists), bool_or(download.event_exists) ) IS NOT NULL AS finished,
-  bool_or(finish_self_print.event_exists) IS NOT NULL AS self_print,
-  bool_or(finish_external.event_exists) IS NOT NULL AS finish_external,
-  bool_or(finish_leo.event_exists) IS NOT NULL AS finish_leo,
+  action.uuid AS action_id,
+  action.events && ARRAY['Finish', 'FinishExternal', 'FinishExternalConfirmed', 'FinishLEO', 'FinishLEOFaxPending', 'FinishLEOFaxSent', 'FinishLEOFaxFailed', 'Download'] AS finished,
+  action.events && ARRAY['FinishPrint'] AS self_print,
+  action.events && ARRAY['FinishExternal', 'FinishExternalConfirmed'] AS finish_external,
+  action.events && ARRAY['FinishLEO', 'FinishLEOFaxPending', 'FinishLEOFaxSent', 'FinishLEOFaxFailed'] AS finish_leo,
```
nickcatal, line 12: Thanks for fixing this

PR #261, opened 2020-07-11
Add the ability to create a test region

benweissmann:
In order to test faxing end-to-end in production (making sure that the prod turnout can talk to the prod fax gateway, etc.) we need to be able to create a region that just sends emails/faxes to us rather than a real LEO.

This adds a Django management command to create a fake region, and adds a "hidden" flag that hides the region from the LEO directory, from the region list when we can't geocode your address, and excludes it from being geocoded to.

Then, to test esign in production, we can turn ON the esign flag, but leave the state-specific fields flag OFF (the frontend won't use the esign flow if state-specific fields are off, so the backend with fall back to self-print because no signature will be attached to the request). Then, we can submit a request directly via cURL to test the fax system end-to-end in the production environment, passing region `999999999` to send the fax/email to us.


app/official/models.py
```diff
@@ -14,12 +14,21 @@ class Meta:
...
 class Region(USVFModel):
     name = models.TextField(null=True)
     municipality = models.TextField(null=True, db_index=True)
     municipality_type = models.TextField(null=True)
     county = models.TextField(null=True, db_index=True)
     state = models.ForeignKey("election.State", null=True, on_delete=models.PROTECT)
+    hidden = models.BooleanField(null=True)
+
+    objects = models.Manager()
+    visible_objects = VisibleRegionManager()
```
nickcatal, line 31: I usually just use `visible`

benweissmann:
The USVF sync doesn't delete regions, but it will delete the address and office records. But that's OK -- i'm thinking that this will be something we run when we need it, not something that we run once and expect to always be there (because we have to re-run it anyway to change which state the test region is in)

PR #277, opened 2020-07-15
map actions to voter registrations

liewegas:
The basic idea here is to map existing app records (verify, register, absentee, reminder) to an active voter registration as accurately as possible.  We can then build on this PR to

- trigger a form on ActionNetwork so we know who is registered (for targeting emails etc)
- see how many of our registrations are actually going through, and how long it takes
- ask users to verify their registration after N weeks if their registration does not appear to have gone through
- prompt users to (verify their registration themselves via their state website and then) request their absentee ballot once their registration is processed

In the future, we may also want to:
- match this to the ``ts.early_vote_and_ballot`` data to notify users of absentee ballot status (when possible)

The best data comes from Alloy: it is updated pretty quickly (for most states) and includes a ``last_updated`` timestamp so we know how fresh it is.  TS is slower and doesn't tell us how stale the data is.  At least one state (WI) has a nice API that lets us look up voters directly (not included here).


app/voter/models.py
```diff
@@ -0,0 +1,124 @@
...
+    city = models.TextField(null=True)
+    zipcode = models.TextField(null=True)
+
+    class Meta:
+        ordering = ["-created_at"]
+
+    def refresh_registration_status(self, registered, register_date, last_updated):
+        if register_date and last_updated and register_date > last_updated:
+            # nonsense!
+            return
```
benweissmann, line 53: When would this happen? Should we warn here?

liewegas:
> Just checking -- this is just adding a background job to create and populate a new `Voter` model, but it's not currently doing anything with that data?

Right.  The idea is to start doing the matching for existing records while we decide exactly how to use it.  If nothing else it would be nice to confirm that registrations are going through in all regions.

I'm considering pulling out the copy of PII in the voter record--or at leaset changing is so that it comes from the original action (+ alloy confirmation) and not the voterfile.. I'm worried about new registrations at a new address vs old TS voterbase records.

app/voter/models.py
liewegas: If alloy has a bad record that shows a registration date after the update date.  Maybe better to warn on the alloy side and assert here.

PR #278, opened 2020-07-16
common/geocode: remove alpha suffix from street number for geocod.io

liewegas:
"40A St Paul St" is parsed as "40 A St".  Work around this by stripping
the 'A' off of the street number.


test_address.py
```diff
@@ -103,7 +103,7 @@
         ("85 Treadwell Hollow Rd", "Williamstown", "MA", "01267", [{'external_id': 431441, 'name': 'Town of Williamstown'}]),
         ("554 Sloan Rd", "Williamstown", "MA", "01267", [{'external_id': 431441, 'name': 'Town of Williamstown'}]),
         ("po box 1", "HOUSATONIC", "MA", "01236", [{'name': 'Town of West Stockbridge', 'external_id': 431427}]),
-        ("69 pleasant st", "gardner", "MA", "01440", [{"name":"Ctiy of Gardner","external_id":431107}]),
+        ("69 pleasant st", "gardner", "MA", "01440", [{"name":"City of Gardner","external_id":431107}]),
```
benweissmann, line 106: Can we add 40A St Paul Street here?

app/common/tests/test_geocode.py
```diff
@@ -73,3 +78,11 @@ def test_well_formed_request(requests_mock):
 )
 def test_al_jefferson_county_bessemer_division(location, inside):
     assert al_jefferson_county_bessemer_division(location) == inside
+
+
+@pytest.mark.parametrize(
+    "before,after", [("40A St Paul St", "40 St Paul St"), ("123 A St", "123 A St"),],
```
benweissmann, line 84: Maybe also a test to make sure that address without a numeric portion aren't altered, and that only the first component is touched -- e.g. `St Paul St` -> `St Paul St`, `St 2A Paul Street` -> `St 2A Paul Street`

PR #280, opened 2020-07-16
actionnetwork improvements

liewegas:
- sync reminders
- use per-action forms
- pass referrer data


liewegas:
https://actionnetwork.org/docs/v2/record_submission_helper references one other field that is not included here:
```
referrer | string |   | The referrer code of this submission. Equivalent to someone taking action with the url argument ?referrer=[your referrer code]. Must be a valid Action Network referrer code. Read-only. Corresponds to the referrers chart in this action's manage page. Action Network-only.
```
I'm not sure if it's applicable here?

app/integration/actionnetwork.py
```diff
@@ -23,13 +25,59 @@ class ActionNetworkError(Exception):
...
+            an_id, va_action = get_form_ids(response.json()["identifiers"])
+            if an_id and va_action:
+                r[va_action] = an_id
+    return r
+
+
+def post_person(info, action, api_key):
     from common.apm import tracer
 
+    forms = setup_action_forms()
```
nickcatal, line 72: We can definitely cache this (for a very very long time) in redis using django's cache framework

app/integration/actionnetwork.py
```diff
@@ -23,13 +25,59 @@ class ActionNetworkError(Exception):
...
+def get_form_ids(ids):
+    an_id = None
+    va_action = None
+    for i in ids:
+        (a, b) = i.split(":")
+        if a == "action_network":
+            an_id = b
+        elif a == "voteamerica":
+            va_action = b
+    return an_id, va_action
```
nickcatal, line 37: can you use vars that are longer than 1 character. this is a bit hard to follow 😄

app/integration/actionnetwork.py
```diff
@@ -23,13 +25,59 @@ class ActionNetworkError(Exception):
...
+    return r
+
+
+def post_person(info, action, api_key):
     from common.apm import tracer
 
+    forms = setup_action_forms()
+    form_id = forms.get(action)
+
     with tracer.trace("an.person.add", service="actionnetwork"):
```
nickcatal, line 75: we're no longer using the `add` endpoint, so likely want to change this

app/integration/actionnetwork.py
```diff
@@ -98,20 +145,29 @@ def _sync_item(item, subscriber_id):
...
+                        "postal_code": item.zipcode,
+                        "country": "US",
+                    },
+                ],
+            },
+            "action_network:referrer_data": {
+                "source": item.source,
+                "website": item.embed_url,
+                "email_referrer": item.email_referrer,
+                "mobile_referrer": item.mobile_referrer,
```
nickcatal, line 167: thanks for this. this is going to be super useful

app/integration/actionnetwork.py
```diff
@@ -23,13 +25,59 @@ class ActionNetworkError(Exception):
...
+    for action in ["lookup", "register", "ballotrequest", "reminderrequest"]:
+        if action not in r:
+            # This code runs once per action, ever.
+            logger.info(f"Creating action form for {action}")
+            response = requests.post(
+                FORM_ENDPOINT,
+                headers={"OSDI-API-Token": settings.ACTIONNETWORK_KEY},
+                json={
+                    "identifiers": [f"voteamerica:{action}"],
+                    "title": action,
```
nickcatal, line 59: We definitely want to prefix the names of these forms with "VoteAmerica" or something so that subscribers know that our tools are the origin of the form when they do targeting

app/integration/actionnetwork.py
```diff
@@ -23,13 +25,59 @@ class ActionNetworkError(Exception):
...
+    for form in response.json()["_embedded"]["osdi:forms"]:
+        an_id, va_action = get_form_ids(form["identifiers"])
+        if an_id and va_action:
+            r[va_action] = an_id
+
+    for action in ["lookup", "register", "ballotrequest", "reminderrequest"]:
+        if action not in r:
+            # This code runs once per action, ever.
+            logger.info(f"Creating action form for {action}")
+            response = requests.post(
```
nickcatal, line 54: def want to wrap this (and the GET above) inside an APM tracer

liewegas:
So the good news is that we can push all the improved referral info to actionnetwork by just blowing away the ``integration_link`` table and repeating the full sync.  The bad/annoying news is that means turning off the ``ACTIONNETWORK_DAILY_SYNC`` temporarily while the full sync runs manually, and then turning it back on afterwards (next day or whatever).

app/integration/actionnetwork.py
```diff
@@ -3,33 +3,99 @@
...
 
-ADD_ENDPOINT = "https://actionnetwork.org/api/v2/people/"
+FORM_ENDPOINT = "https://actionnetwork.org/api/v2/forms/"
+ADD_ENDPOINT = "https://actionnetwork.org/api/v2/forms/{form_id}/submissions/"
 
 logger = logging.getLogger("integration")
 
+ACTIONS = ["lookup", "register", "ballotrequest", "reminderrequest"]
+
+CACHE_KEY = "lookup", "register", "ballotrequest", "reminderrequest"
```
nickcatal, line 26: Wouldn't we need separate forms and cache keys for each subscriber?

app/integration/actionnetwork.py
```diff
@@ -3,33 +3,99 @@
...
+                        "identifiers": [f"voteamerica:{action}"],
+                        "title": f"VoteAmerica {action}",
+                        "origin_system": "voteamerica",
+                    },
+                )
+                an_id, va_action = get_form_ids(response.json()["identifiers"])
+                if an_id and va_action:
+                    forms[va_action] = an_id
+
+    cache.set(CACHE_KEY, forms)
```
nickcatal, line 82: By default django only caches for 5 minutes https://docs.djangoproject.com/en/3.0/ref/settings/#timeout

We likely want to set this timeout to be something much longer (and allow it to be adjusted in settings.) I'd imagine at a minimum we'd want it to be cached for a day?

app/integration/actionnetwork.py
```diff
@@ -3,33 +3,99 @@
...
+        if action not in forms:
+            # This code runs once per action, ever.
+            logger.info(f"Creating action form for {action}")
+            with tracer.trace("an.form.create", service="actionnetwork"):
+                response = requests.post(
+                    FORM_ENDPOINT,
+                    headers={"OSDI-API-Token": settings.ACTIONNETWORK_KEY},
+                    json={
+                        "identifiers": [f"voteamerica:{action}"],
+                        "title": f"VoteAmerica {action}",
```
nickcatal, line 74: This is random, but this will render to `VoteAmerica registration`, since we're lowercasing our action names. Can we make it `VoteAmerica BallotRequest` or `VoteAmerica Registration` (the full camelcase class name)

app/integration/actionnetwork.py
```diff
@@ -98,20 +162,29 @@ def _sync_item(item, subscriber_id):
...
+                        "address_lines": [item.address1, item.address2],
+                        "locality": item.city,
+                        "region": item.state_id,
+                        "postal_code": item.zipcode,
+                        "country": "US",
+                    },
+                ],
+            },
+            "action_network:referrer_data": {
+                "source": item.source,
```
nickcatal, line 181: Does it make sense to have a default source, maybe `voteamerica_{actionname}` or something? Just thinking, I dunno here.

app/integration/actionnetwork.py
```diff
@@ -3,33 +3,99 @@
...
+
+    forms = {}
+    with tracer.trace("an.form", service="actionnetwork"):
+        response = requests.get(
+            FORM_ENDPOINT, headers={"OSDI-API-Token": settings.ACTIONNETWORK_KEY},
+        )
+        for form in response.json()["_embedded"]["osdi:forms"]:
+            an_id, va_action = get_form_ids(form["identifiers"])
+            if an_id and va_action:
+                forms[va_action] = an_id
```
nickcatal, line 62: What happens if a subscriber has hundreds of forms and the ActionNetwork response paginates?

PR #286, opened 2020-07-17
Enable esign for KS

camdenlee:
- Only listed November election. By the time this is enabled, it will be too close to the state-wide (non-presidential) primary election (2020-08-04)
- Form requires a state ID number or you can copy different ID types. Since we do not have the ability to scan alternate ID types, I've set it to require state ID number
- The state wants the state and county from where you submit the form. If you submit from vacation in Hawaii, they request that. We've used the state and county of the standard address.

# default state 

Screen Shot 2020-08-12 at 4 19 28 PM

# filled in

Screen Shot 2020-08-12 at 4 19 53 PM

# signature

Screen Shot 2020-07-17 at 11 42 40 AM

# PDF

Screen Shot 2020-07-17 at 11 37 14 AM


benweissmann:
Talked to Debra about the ID thing -- the plan we came up with is to have a radio group, and if you pick "no" then we don't let you esign. I pushed a commit that adds that flag to the YAML, so we just need to add support on the frontend: https://app.clubhouse.io/turnout2020/story/1633/support-blocks-esign-in-state-specific-fields (I think this should be pretty straightforward, but if it's not let me know and we can come up with a different plan)

camdenlee:
@benweissmann this should be ready for review!

PR #287, opened 2020-07-17
Replace Absentee Overrides admin panel with a VBM E-Sign Dashboard

benweissmann:
Replaces the basic dashboard for creating/updating/deleting overrides, with a new dashboard that provides lots more visibility into the esign program, as well as the ability to manage overrides.


docker-compose.yaml
```diff
@@ -29,7 +29,7 @@ services:
     entrypoint: wait-for-it postgres:5432 redis:6379 --
     command: python manage.py runserver 0.0.0.0:8000
     volumes:
-      - ./app:/app
+      - ./app:/app:cached
```
benweissmann, line 32: This uses Docker for Mac caching when available to speed up disk access

app/event_tracking/migrations/0004_auto_20200716_1604.py
```diff
@@ -0,0 +1,18 @@
...
+        ('event_tracking', '0003_event_type_db_index'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='event',
+            name='created_at',
+            field=models.DateTimeField(auto_now_add=True, db_index=True),
+        ),
+    ]
```
nickcatal, line 18: Random request, can you give this migration a name other than `auto_20200716_1604.py`

app/event_tracking/models.py
```diff
@@ -6,6 +6,9 @@
 
 
 class Event(UUIDModel, TimestampModel):
+    # Override TimestampModel to index this field
+    created_at = models.DateTimeField(auto_now_add=True, db_index=True)
```
nickcatal, line 10: smart

app/absentee/migrations/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+	  ballotrequest.region_id AS external_id,
+	  COUNT(DISTINCT ballotrequest.action_id) FILTER (WHERE event.event_type = 'FinishLEO') AS emails_sent,
+	  COUNT(DISTINCT ballotrequest.action_id) FILTER (WHERE event.event_type = 'FinishLEOFaxPending') AS faxes_sent
+	FROM absentee_ballotrequest ballotrequest
+	JOIN event_tracking_event event ON event.action_id = ballotrequest.action_id
+	WHERE event.created_at > NOW() - interval '1 days'
+	AND event.event_type IN ('FinishLEO', 'FinishLEOFaxPending')
+	GROUP BY ballotrequest.region_id
+) stats_1d USING (external_id);
+"""
```
nickcatal, line 164: I've been trying to prevent doing counts like these in real-time in the admin. It can slow django down quite a bit when there is significant amounts of data to be displayed (like there will be later in the cycle.) That's why things like stats are done using a background task.

What is this going to perform like when we have thousands and thousands of submissions?

app/absentee/dashboard/forms.py
```diff
@@ -0,0 +1,96 @@
...
+    class Meta(object):
+        model = LeoContactOverride
+        fields: List[str] = []
+
+
+class LeoContactOverrideManageCreateForm(LeoContactOverrideManageUpdateForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["region"].widget = forms.HiddenInput()
+        self.fields["region"].initial = kwargs["region_id"]
```
nickcatal, line 92: As a general philosophy I dislike putting primary keys in hidden fields except in very limited circumstances.

Is there some way you can just set the region id right before the view saves the model?

app/absentee/dashboard/manage_views.py
```diff
@@ -0,0 +1,193 @@
...
+    context_object_name = "states"
+    template_name = "absentee/dashboard/state_dashboard.html"
+
+    def get_queryset(self):
+        queryset = super().get_queryset()
+
+        esign_states = [
+            state_code
+            for (state_code, state_name) in STATES
+            if flag_enabled_for_state("vbm_esign", state_code)
```
nickcatal, line 43: Wouldn't we want to expose all the e-sign states regardless of if they're active inside optimizely or not, so that we can do work setting up overrides before we switch over optimizely (and thus enable the site in the frontend)

app/absentee/dashboard/manage_views.py
```diff
@@ -0,0 +1,193 @@
...
+        context["region"] = Region.objects.get(pk=self.kwargs["pk"])
+        return context
+
+    def get_form_kwargs(self):
+        return {
+            "region_id": self.kwargs["pk"],
+            "usvf_data": get_absentee_contact_info(
+                self.kwargs["pk"], skip_overrides=True
+            ),
+            **super().get_form_kwargs(),
```
nickcatal, line 186: the `**super().get_form_kwargs()` style isn't something we've done elsewhere, but I think we should start doing it.

app/absentee/dashboard/manage_views.py
benweissmann: The email data from USVF is pretty complete -- I think we're going to be using this just to handle overrides which we won't know about until post-launch (and if we know pre-launch we can just tell Susan to update the USVF data). I'd rather keep this to just active esign states to keep it simpler.

app/absentee/migrations/__init__.py
benweissmann: So right now, running this query (without filtering by state, and without adding the new indices) takes 11ms. I think that given that we only run this filtered to one state, and we're adding new indices to optimize the counts (including a compound index on `created_at, event_type`, which means we can count directly from the index without loading the pages), we're probably good to handle a few orders of magnitude of scale-up.

We could pre-compute this or cache it, but I think that's something we can do if this page slows down.

PR #289, opened 2020-07-19
Add Maryland pdf

amyrlam:
Story: https://app.clubhouse.io/turnout2020/story/1587/maryland-vbm-pdf-app


amyrlam:
@benweissmann @camdenlee @nickcatal Not sure how to get past this error yet, let me know if anything jumps out at you! 

(1)
When I run `make testpdf STATE=MD` locally, I get this error:

```
E               AssertionError: Mismatch for field phone: expected (617) 555-1234, got (6
E               assert '(6' == '(617) 555-1234'
E                 - (617) 555-1234
E                 + (6

absentee/tests/test_pdf.py:96: AssertionError
```

(2)
I was hoping to debug (1) ^^ from the pdf emailed to me. But when I go to that I get a 404?

I have one terminal running `make up`.

bug

(BTW I'm just working on this now cause in the quarlife, evenings are less busy!)

amyrlam:
One more question: 

(3)
I didn't do the `auto_fields` because the field for "Today's Date" was separated with pre-printed slashes. Possible to extract `todays_month` / `todays_day` / `todays_year` for `auto_fields`, or is it better to just do manually?

benweissmann:
(1) So what's happening here is that as part of our tests, we take the test data, fill the form with it, and then read back the PDF and make sure it matches the test data. In this case, the phone field is set up with a "comb" of two characters (comb is just another form of length limit, designed to also space out the text):

Screen Shot 2020-07-19 at 10 50 03 AM

Which is truncating the phone number to just the first two characters.

It looks like the "fax" field any maybe others also have this option -- disabling the "comb" option should fix this.

(2) @nickcatal can probably help more here -- I use a slightly different setup for my dev environment -- but take a look in the `app/uploads` folder in your Turnout repo -- the file should be in there so you can grab it from that folder.

(3) You can just edit the form to replace that with a single `todays_date` field.

amyrlam:
@benweissmann thank you! 💯

I think I solved everything on (1), (2) and (3). Here is the pdf:

[200719.vyeN7g.md-weissmann-ballotrequest.pdf](https://github.com/vote/turnout/files/4945350/200719.vyeN7g.md-weissmann-ballotrequest.pdf)

I made a small tweak to the Notion instructions.

amyrlam:
Thanks for the feedback. Pushed another commit and here's the new pdf, let me know what you think: [200720.5h49Bag4B7.md-weissmann-ballotrequest.pdf](https://github.com/vote/turnout/files/4949779/200720.5h49Bag4B7.md-weissmann-ballotrequest.pdf)

I just noticed the note about the Miami cover sheet in the Notion, apologies.

PR #293, opened 2020-07-20
scripts/remote_run.sh: fix ecr login

liewegas:
The old version fails on my box, this version works (and
matches the docs).

https://docs.aws.amazon.com/cli/latest/reference/ecr/get-login-password.html


scripts/remote_run.sh
```diff
@@ -11,7 +11,7 @@ ACCOUNT_ID=$(aws sts get-caller-identity | jq -r ".Account")
 if [ $1 ]; then
 
 echo "Logging into ECR"
-eval $(aws ecr get-login --no-include-email --region $REGION)
+aws ecr get-login-password --region $REGION | docker login --username AWS --password-stdin $ACCOUNT_ID.dkr.ecr.$REGION.amazonaws.com
```
benweissmann, line 14: This requires the AWS CLI v2 (i think -- I had to upgrade before `get-login-password` would work). Can you add a check for `aws --version | grep 'aws-cli/2'` and bail if the aws cli is out of date?

Also, this file is auto-generated from here: https://github.com/vote/turnout/blob/master/ecs/template/remote_run.jsonnet

liewegas: Good call--current version works with both v1 and v2

PR #294, opened 2020-07-20
actionnetwork: separate action forms for prod and staging/dev/local

liewegas:
The form ids are global, so we need to have distinct IDs across prod
and staging/dev/local (which share the 'dev' API key).


app/integration/actionnetwork.py
```diff
@@ -78,33 +78,36 @@ def setup_action_forms(subscriber_id):
 
     logger.info(f"Fetching forms from ActionNetwork for subscriber {subscriber_id}")
     api_key = get_api_key(subscriber_id)
+    # we have a prod API key and a staging one (shared by staging, dev, local)
+    prefix = "prod" if settings.ENV == "prod" else "staging"
```
nickcatal, line 82: Can we just store this setting in the settings file/ECS?

liewegas: Yep, updated, and AWS vars created.

PR #300, opened 2020-07-22
register: fix region_ids

liewegas:


app/register/api_views.py
```diff
@@ -42,13 +42,12 @@ def after_create_or_update(self, registration):
...
-                queryset = Region.visible.filter(state__code="PA").order_by("name")
                 regions = []
-                for region in queryset:
-                    regions.append(region.external_id)
+                for region in Region.visible.filter(state__code="PA").order_by("name"):
+                    regions.append(region)
             if not registration.state_api_result:
                 registration.state_api_result = {}
-            registration.state_api_result["regions"] = regions
+            registration.state_api_result["regions"] = [region.external_id for region in regions]
```
benweissmann, line 50: I think we're going to want an array of `{ name: str, external_id: str }` -- same as we return for absentee geocoding: https://github.com/vote/turnout/blob/master/app/absentee/api_views.py#L146 (because the frontend needs to show the user a list of regions, so it needs the names)

app/register/tests/test_api_views.py
```diff
@@ -417,20 +417,20 @@ def test_bad_update_status():
 @pytest.fixture
 def mock_region(mocker):
     class FakeRegion(object):
-        def __init__(self, name):
+        def __init__(self, name, external_id):
```
benweissmann, line 420: Can we add a couple more geocoding tests? To exercise the codepaths where there are a few regions, or we fall back to all regions?

PR #301, opened 2020-07-22
add 'make syncusvf' and allow disabling geocoding

liewegas:
Add USVOTEFOUNDATION_GEOCODE=False to .env and `make syncusvf`.


app/turnout/settings/core.py
```diff
@@ -528,6 +528,8 @@
 #### ELECTION OFFICIALS CONFIGURATION
 
 USVOTEFOUNDATION_KEY = env.str("USVOTEFOUNDATION_KEY", default=None)
+USVOTEFOUNDATION_GEOCODE = env.bool("USVOTEFOUNDATION_GEOCODE", default=True)
```
nickcatal, line 531: Can we default this to `False` so that people who forget to add this to their `.env` aren't in for a nasty surprise?

liewegas: Yep!  Updated, and added the var to AWS.

PR #309, opened 2020-07-23
registration: log event on completion via external api; send email

liewegas:


liewegas:
There was no actual migration... it looks like the EnumField is just a text field in postgres?

app/action/migrations/__init__.py
```diff
@@ -6,7 +6,7 @@
 CREATE OR REPLACE VIEW action_actiondetails AS
 SELECT
   action.uuid AS action_id,
-  action.events && ARRAY['Finish', 'FinishExternal', 'FinishExternalConfirmed', 'FinishLEO', 'FinishLEOFaxPending', 'FinishLEOFaxSent', 'FinishLEOFaxFailed', 'Download'] AS finished,
+  action.events && ARRAY['Finish', 'FinishExternal', 'FinishExternalAPI', 'FinishExternalConfirmed', 'FinishLEO', 'FinishLEOFaxPending', 'FinishLEOFaxSent', 'FinishLEOFaxFailed', 'Download'] AS finished,
```
benweissmann, line 9: You need to manually create a migration for this change -- a la https://github.com/vote/turnout/blob/master/app/action/migrations/0005_refresh_action_view.py

app/register/api_views.py
```diff
@@ -184,6 +188,9 @@ def combine_addr(a, b):
                 "signature_source": res.signature_source,
             }
             registration.status = TurnoutActionStatus.PENDING
+            registration.action.track_event(EventType.FINISH_EXTERNAL_API)
+            send_registration_notification.delay(registration.pk)
```
benweissmann, line 192: I don't think this notification is going to work -- this emails the voter with a copy of their PDF and tells them to print and mail it. We'll need a new transactional email for this.

liewegas: :facepalm: right

liewegas:
The confirmation email needs a close look--I just guessed at what we should be including there.

app/register/templates/register/email/state_confirmation.html
```diff
@@ -0,0 +1,26 @@
+{% extends "email/base.html" %}
+{% load markdownify %}
+{% block content %}
+
+
+{{ recipient.first_name }},<br><br>
+Your voter registration has been submitted!
+<br><br>
+
+You are not registered to vote until your state processes this form.  You should receive written confirmation within 1-2 weeks.
```
benweissmann, line 10: Is there somewhere that says that the state definitely sends a written confirmation?

app/register/templates/register/email/state_confirmation.html
```diff
@@ -0,0 +1,26 @@
...
+If not, you might want to <a href="https://www.voteamerica.com/local-election-offices/{{registration.state.code}}/">call your local election office</a>{% if state_info.external_tool_verify_status %} or use the <a href="{{ state_info.external_tool_verify_status }}">State of {{ registration.state }} website</a>{% endif %} to confirm that you are now registered to vote.<br><br>
```
benweissmann, line 11: I'd use stronger language here -- "If you don't receive written confirmation within 2 weeks, you should..."

app/register/tests/test_api_views.py
```diff
@@ -133,6 +137,7 @@
 @pytest.fixture()
 def submission_task_patch(mocker):
     mocker.patch("register.api_views.send_welcome_sms")
+    mocker.patch("register.api_views.send_registration_state_confirmation")
```
benweissmann, line 140: Should add a couple assertions to make sure this gets called only on success

PR #314, opened 2020-07-27
Convert SmallUUID to UUID

nickcatal:
So that it's vastly easier to link URLs in the admin (and elsewhere) to records in the database, switch our usage of SmallUUIDs to UUIDs.


app/accounts/views.py
```diff
@@ -19,6 +21,15 @@ class InviteConsumeFormView(UUIDSlugMixin, SingleObjectMixin, FormView):
     context_object_name = "invite"
     success_url = "/"
 
+    def dispatch(self, request, *args, **kwargs):
+        try:
+            kwargs["slug"] = SmallUUID(kwargs["slug"]).hex_grouped
+            self.kwargs["slug"] = kwargs["slug"]
+            statsd.increment("turnout.accounts.invite_smalluuid_key_usage")
+        except Exception:
```
benweissmann, line 29: Can you be more specific about this Exception? Like is this supposed to fail if `kwargs["slug"]` isn't a SmallUUID? I'm worried that this will end up swallowing errors from, e.g., `statsd.increment`.

app/storage/public_views.py
```diff
@@ -15,6 +16,15 @@
 class DownloadFileView(UUIDSlugMixin, SingleObjectMixin, RedirectView):
     model = StorageItem
 
+    def dispatch(self, request, *args, **kwargs):
+        try:
+            kwargs["pk"] = SmallUUID(kwargs["pk"]).hex_grouped
+            self.kwargs["pk"] = kwargs["pk"]
+            statsd.increment("turnout.storage.download_smalluuid_key_usage")
+        except Exception:
```
benweissmann, line 24: Ditto on being more specific

liewegas:
Missed one ``hex_grouped`` instance in ``integration/actionnetwork.py``

app/absentee/migrations/0020_add_dashboard_views.py
```diff
@@ -13,8 +13,8 @@ class Migration(migrations.Migration):
     dependencies = [
         ("absentee", "0019_actionnetwork_tracking"),
         ("official", "0006_region_hidden"),
-        ("election", "0008_auto_20200629_2252"),
-        ("event_tracking", "0004_add_event_type_created_at_index"),
+        ("election", "0001_initial"),
+        ("event_tracking", "0002_enumchartotextfield"),
```
liewegas, line 17: Why change the dependencies on this old migration?

nickcatal: It was causing a migration order problem that made it impossible to create new migrations.

app/accounts/views.py
nickcatal: Done

PR #316, opened 2020-07-28
smsbot: add ability to proxy webhook (+ sniff STOP/JOIN etc)

liewegas:
This will let us keep twilio pointed at us when we switch to ActionNetwork and keep a master version of our opt-out list.


liewegas:
We don't want to intercept the webhook.

PR #317, opened 2020-07-28
election: add endpoint to redirect to state tools

liewegas:
This will let us put links in emails that send users directly to the state
tools.  This is desirable for things like having the user verify their voter
registration went through--we want them to use the authoritative state tool
and not our laggy check.  And since they're already a user there is no
point in re-collecting their info.

If there is no state tool, redirect to our internal tool.

Note: for CO we are passing only 1 of the 2 tracking args.. is this okay, or should we pass none?

### DIFF
--- a/app/election/api_urls.py
+++ b/app/election/api_urls.py
@@ -1,10 +1,16 @@
+from django.urls import path
 from rest_framework import routers
 
-from .api_views import StateFieldsViewSet, StateViewSet
+from .api_views import StateExternalToolRedirectView, StateFieldsViewSet, StateViewSet
 
 router = routers.SimpleRouter()
 router.register(r"state", StateViewSet)
 router.register(r"field", StateFieldsViewSet)
 
 app_name = "api_election"
-urlpatterns = router.urls
+urlpatterns = router.urls + [
+    path(
+        "state-redirect/<str:state>/<str:slug>/",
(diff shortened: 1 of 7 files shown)


benweissmann:
Passing just one arg for CO is fine -- you should chat with @nickcatal about having Cloudflare cache this endpoint so we don't get hammered when we send a big email.

nickcatal:
This isn't really a REST-ful API, so I don't think we'd really want it at api.voteamerica.com/v1/

Is there no other way to handle this? Perhaps a s3 bucket with redirect objects or something?

liewegas:
> This isn't really a REST-ful API, so I don't think we'd really want it at api.voteamerica.com/v1/
> 
> Is there no other way to handle this? Perhaps a s3 bucket with redirect objects or something?

Redirect objects in S3 would work, and would certainly incur less load (we'd just update them daily or whatever).  The real limitation is that actionnetwork's email fields aren't super sophisticated, so we can substitute in the state or misc personal info but not second-order fields custom per-state info.  https://help.actionnetwork.org/hc/en-us/articles/203846565-Using-clips-to-personalize-emails-and-mobile-messages

One benefit of using an API like this is we could track how many/which users follow the link.  I'm not sure if that's useful information or not.

Just let me know which direction you'd like me to take!

PR #327, opened 2020-07-31
Reduce size of ECS task definitions

nickcatal:
ECS task definitions max out at 65536 bytes. Ours were at 65937. By removing a ton of unnecessary space while we wait to break out task definitions we won't run into this problem.


benweissmann:
Instead of doing this manually, can we just run this files through [jq](https://nelsonslog.wordpress.com/2016/08/19/using-jq-to-minify-json/) in the `build.sh` script?

nickcatal:
@benweissmann was in the process of doing exactly that

PR #339, opened 2020-08-03
remove secrets from beat container; re-add uptime secrets

liewegas:


app/leouptime/api_views.py
```diff
@@ -1,17 +1,24 @@
 from rest_framework import generics
+from rest_framework.pagination import LimitOffsetPagination
```
benweissmann, line 2: looks like you might need to rebase this on master?

PR #345, opened 2020-08-04
smsbot: track SMS message delivery

liewegas:
This serves two purposes: (1) we have delivery feedback, so we can stop texting bad numbers (TODO), and (2) we prevent the actionnetwork delivery webhook from getting called for messages that we send.


app/smsbot/models.py
```diff
@@ -1,32 +1,67 @@
...
+            return
+        msg = SMSMessage.objects.create(
+            phone=str(self.phone), direction=MessageDirectionType.OUT, message=text,
+        )
+        r = twilio_client.messages.create(
+            to=str(self.phone),
+            messaging_service_sid=settings.TWILIO_MESSAGING_SERVICE_SID,
+            body=text,
+            status_callback=msg.delivery_status_webhook(),
+        )
```
nickcatal, line 67: Can you wrap this in an APM trace

PR #350, opened 2020-08-05
storage: purge old pdfs, signatures

liewegas:


app/storage/tests/test_api_views.py
```diff
@@ -66,3 +68,40 @@ def test_reset_no_such_id(client):
     resp = client.post(STORAGE_RESET_URL, {"id": uuid4()})
 
     assert resp.status_code == 404
+
+
+@pytest.mark.django_db
+def test_expired_ballot_request_form(client):
```
benweissmann, line 74: Can you also test purging SecureUploadItems, and also expand the tests to make sure we *don't* purchase newer items?

app/storage/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+from django.conf import settings
+
+from .models import SecureUploadItem, StorageItem
+
+logger = logging.getLogger("storage")
+
+
+@shared_task
+def purge_old_storage():
+    if settings.FILE_PURGE_HOURS:
```
benweissmann, line 14: Maybe also check `FILE_PURGE_HOURS >= 1` to make it explicit that we can set this to 0 or -1 to disable the behavior

app/turnout/settings/core.py
```diff
@@ -352,6 +352,17 @@
     "FILE_TOKEN_RESET_URL", default="http://localhost:8000/download/?id={item_id}"
 )
 
+FILE_PURGE_HOURS = env.int("FILE_PURGE_HOURS", default=None)
+FILE_TOKEN_PURGED_URL = env.str(
+    "FILE_TOKEN_PURGED_URL",
+    default="http://localhost:8000/download-purged/?id={item_id}&filetype={filetype}",
+)
+if FILE_PURGE_HOURS:
```
benweissmann, line 360: Let's go ahead an add this to the ECS task definition, but set it to `-1` in production for now (we can set it to 2 months in dev/staging)

app/storage/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+        now = datetime.datetime.now(tz=datetime.timezone.utc)
+        purge_before = now - datetime.timedelta(hours=settings.FILE_PURGE_HOURS)
+        for item in StorageItem.objects.filter(
```
benweissmann, line 17: Should filter this by type -- only RegForm and AbsForm should be purged

app/storage/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+            purged__isnull=True, created_at__lt=purge_before
+        ):
+            logger.info(f"Purging {item}")
+            item.purged = now
+            item.file.delete(save=True)
+
+        for item in SecureUploadItem.objects.filter(
+            purged__isnull=True, created_at__lt=purge_before
```
benweissmann, line 25: Ditto on filtering by type -- right now the only type is Signature but we should be explicit here so that if we use SecureUploadItem for stuff in the future it doesn't get purged by default

PR #351, opened 2020-08-05
actionnetwork: allow subscribers to configure api key in settings; fix bug with form id

liewegas:


app/multi_tenant/manage_views.py
```diff
@@ -44,9 +45,45 @@ class SubscriberUpdateSettingsView(
...
     def get_object(self):
         return self.subscriber
 
+    def get_context_data(self, **kwargs):
+        data = super().get_context_data(**kwargs)
+
+        old_key = self.get_actionnetwork_key(data["form"].instance)
+        if old_key:
+            data["form"].initial["sync_actionnetwork"] = True
+            data["form"].initial["actionnetwork_api_key"] = old_key.value
```
benweissmann, line 64: Does this cause the actionnetwork key to be shown in the admin UI? I think we should treat this like a password -- you can enter a new value, but you can't see the current value.

liewegas: Yep, fixed!

liewegas:
I couldn't figure out how to make the API key input box indented so that it is grouped "beneath" the sync checkbox...

benweissmann:
Hiding the API key isn't quite working -- it's an `input type="password"` so it shows up as dots in the UI, but if you look at the source, the API key is in the HTML. I think when we're rendering the HTML, we should render it as `<input type="password" value="****" />` and then when we're processing the form submission, we can just ignore that field if the value is all `*`s.

For indenting -- that input has the id `div_id_actionnetwork_api_key`, so you can use this css:

```
#div_id_actionnetwork_api_key {
  margin-left: 50px;
}
```

You can stick that at the bottom of this file: https://github.com/vote/turnout/blob/master/app/assets/scss/manage/_widgets.scss -- but our build pipeline has a bug where changes to scss includes (files starting with `_`) don't always trigger a rebuild, so you might have to make a change to this file to get it to rebuild: https://github.com/vote/turnout/blob/master/app/assets/scss/app.scss

Last thing -- right now, you can enter an API key without enabling ActionNetwork sync. Is there ever a reason to let users do that? Or can we require the checkbox to be checked if you enter an API key?

PR #352, opened 2020-08-06
Simplify SMS opt-in behavior selection

liewegas:


app/multi_tenant/models.py
```diff
@@ -89,6 +89,34 @@ def disclaimer(self):
...
+            return enums.SubscriberSMSOption.NONE
+        if not self.sms_checkbox:
+            return enums.SubscriberSMSOption.AUTO_OPT_IN
+        if self.sms_checkbox_default:
+            return enums.SubscriberSMSOption.BOX_CHECKED
+        else:
+            return enums.SubscriberSMSOption.BOX_UNCHECKED
+
+    def set_sms_mode(self, opt):
+        if opt == str(enums.SubscriberSMSOption.NONE):
```
liewegas, line 103: I couldn't get the str -> enum conversion to work.  The value that ends up in the forms in CamelCase with spaces not underscores (e.g., ``Box Checked``), and the usual ``MyEnumType[string]`` doesn't work with those values.  Couldn't unravel why it was behaving that way so I compared as strings here... I think it's something in django since the enum looks like this:
```
In [40]: common.enums.SubscriberSMSOption.__dict__                                                                                                            
Out[40]: 
mappingproxy({'_generate_next_value_': <function enum.Enum._generate_next_value_(name, start, count, last_values)>,
              '__module__': 'common.enums',
              '__doc__': 'An enumeration.',
              '_member_names_': ['NONE',
               'AUTO_OPT_IN',
               'BOX_UNCHECKED',
               'BOX_CHECKED'],
              '_member_map_': OrderedDict([('NONE',
                            <SubscriberSMSOption.NONE: 'none'>),
                           ('AUTO_OPT_IN',
                            <SubscriberSMSOption.AUTO_OPT_IN: 'auto_opt_in'>),
                           ('BOX_UNCHECKED',
                            <SubscriberSMSOption.BOX_UNCHECKED: 'box_unchecked'>),
                           ('BOX_CHECKED',
                            <SubscriberSMSOption.BOX_CHECKED: 'box_checked'>)]),
              '_member_type_': object,
              '_value2member_map_': {'none': <SubscriberSMSOption.NONE: 'none'>,
               'auto_opt_in': <SubscriberSMSOption.AUTO_OPT_IN: 'auto_opt_in'>,
               'box_unchecked': <SubscriberSMSOption.BOX_UNCHECKED: 'box_unchecked'>,
               'box_checked': <SubscriberSMSOption.BOX_CHECKED: 'box_checked'>},
              'NONE': <SubscriberSMSOption.NONE: 'none'>,
              'AUTO_OPT_IN': <SubscriberSMSOption.AUTO_OPT_IN: 'auto_opt_in'>,
              'BOX_UNCHECKED': <SubscriberSMSOption.BOX_UNCHECKED: 'box_unchecked'>,
              'BOX_CHECKED': <SubscriberSMSOption.BOX_CHECKED: 'box_checked'>,
              '__new__': <function enum.Enum.__new__(cls, value)>})
```

app/subscription/migrations/0005_interest_sms_mode.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('subscription', '0004_smalluuid_to_uuid'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='interest',
+            name='sms_mode',
+            field=common.fields.TurnoutEnumField(default='box_checked', enum=common.enums.SubscriberSMSOption, null=True),
```
benweissmann, line 18: This should default to `box_unchecked`, which is the current default (we don't want to use a mode that is possibly-illegal for an org without them explicitly choosing it)

liewegas: Should the model be the same, so that it's the default checked box on the subscribe signup form?

liewegas: nm, the model default doesn't affect the radio behavior, but it's required, so they have to choose something.

liewegas:
@benweissmann any final concerns?

PR #353, opened 2020-08-06
Move PDF generation to lambda

benweissmann:
- I extracted PDFTemplate and PyPDFTK to a shared library on pypi: https://github.com/vote/pdf_template
- I created a tiny lambda function that just runs PDFTemplate.fill: https://github.com/vote/pdf_filler
- I added a configuration option to call out to that lambda instead of running PDF generation inline in the app server
- I compressed a few of our PDFs (using the Adobe Acrobat "reduce file size" option) to make passing them around cheaper
- I moved PDF generation out of background tasks and into the foreground request/response cycle

This gets us a couple things:
- Most importantly, it gets state_id_number out of the task queue
- It offloads our one CPU-intensive thing (PDF generation) to Lambda, where it can scale elastically
- It isolates our PDF generation so that a vulnerability in pdftk doesn't really matter -- the lambda is completely unprivileged (it uses presigned s3 URLs for all s3 operations) and runs outside our VPC
- It gives the user better feedback -- if PDF generation fails, we can show a an error message on the frontend


app/common/aws.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+sqs_client = boto3.client(
+    "sqs",
+    config=Config(retries={"max_attempts": 10, "mode": "standard"}),
+    **aws_creds()
+)
+
+
+s3_client = boto3.client("s3", **aws_creds())
```
liewegas, line 26: app/election/external_redirects.py should probably be switched to make use of this

app/common/pdf/pdftemplate.py
```diff
@@ -1,150 +1,119 @@
...
+            "utf-8",
+        )
+
+    with tracer.trace("common.pdf.pdftemplate.call_lambda"):
+        response = lambda_client.invoke(
+            FunctionName="pdf-filler-local-fill",
+            InvocationType="RequestResponse",
+            LogType="Tail",
+            Payload=lambda_payload,
+        )
```
liewegas, line 78: Does this need a timeout?

benweissmann: The lambda function itself has a timeout, so I don't think so.

PR #355, opened 2020-08-06
Update copy in subscriber emails

benweissmann:
Copy is from this Doc that debra edited: https://docs.google.com/document/d/11qL7hxMnlHBIbcSCa1nJTGLomlMTKoObaOAMiXBTJIc/edit

https://app.clubhouse.io/turnout2020/story/1881/update-the-copy-on-the-subscriber-emails


app/multi_tenant/templates/multi_tenant/email/new_invite.html
```diff
@@ -3,7 +3,7 @@
 
 
 Hello,<br><br>
-You've been invited to join {{ subscriber }}'s subscription on the VoteAmerica toolset.
+You've been invited to join the {{ subscriber }} team in the VoteAmerica toolset.
```
camdenlee, line 6: Is there a reason we use `{{ subscriber }}` here and `{{ subscriber_name }}` in welcome.html? Looks like it's defined in the `compile_email` `context`.  It seems like that it was setup as expected, but just wanted a quick gut check on this.

app/subscription/templates/subscription/emails/welcome.html
```diff
@@ -3,33 +3,32 @@
...
+account in order to access the VoteAmerica admin portal.<br><br>
 
-The following are important links to be aware of:<br><br>
+The admin portal portal allows you to do the following:<br>
+<ul>
+<li>Access your custom embed code</li>
+<li>View usage stats</li>
+<li>Export a CSV of your voter data</li>
+<li>Enable or disable a SMS disclaimer checkbox</li>
+<li>Sync to a 3rd party system, such as Bluelink.</li>
```
camdenlee, line 18: remove `.`

app/multi_tenant/templates/multi_tenant/email/new_invite.html
benweissmann: oh whoops, i didn't realize i could pass in just the whole `subscriber` object. Good catch, fixed!

PR #363, opened 2020-08-09
Esign Application PDF Viewing

benweissmann:
To help the program team respond to LEO feedback, this adds a new view to the esign dashboard where specific users can view "previews" of the submitted applications. These previews are generated on-demand and do not have the `state_id_number` on them.

This also reworks permissions a little bit -- there is a new `is_esign_admin` flag on users that controls access to the esign dashboard (previously, the esign dashboard was controlled by access to `is_election_admin`, but because this new functionality is extremely sensitive, I thought it made sense to make it a separate permission.


benweissmann:
Good call -- i'll add a bit of explanatory text

PR #365, opened 2020-08-10
Omit territories from admin tool and API responses

benweissmann:


app/election/migrations/0012_add_territory_field.py
```diff
@@ -0,0 +1,98 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name="state",
+            name="territory",
+            field=models.BooleanField(default=False, null=True),
+        ),
+        migrations.RunPython(mark_territories, lambda x, y: None),
+        migrations.AlterField(
+            model_name="state",
```
liewegas, line 30: I don't understand why this alterfield is there, but it seems harmless

benweissmann: I forgot to run makemigrations before, so it's from the previous change that added additional options

PR #371, opened 2020-08-11
smsbot: fix for remindres, absentee

liewegas:
also fix actionnetwork sync for absentee


app/absentee/tasks.py
```diff
@@ -11,21 +11,15 @@
...
 
-# DEPRECATED
-# TODO: Delete. Leaving this in so that in-flight tasks don't fail during the
-# deploy.
+
 @shared_task
-@statsd.timed("turnout.absentee.process_ballotrequest_submission")
-def process_ballotrequest_submission(
-    ballotrequest_pk: str, state_id_number: str, is_18_or_over: bool
-) -> None:
```
nickcatal, line 21: Are we actively triggering this anywhere currently?

Because of the way we deploy we can't remove tasks like this. We still need one called `process_ballotrequest_submission` that accepts 3 arguments, even if it does nothing but return None

But also I know that SMS isn't turned on, so this might be moot since we're not using this task anywhere

app/reminder/api_views.py
```diff
@@ -29,7 +28,6 @@ def create(self, request, *args, **kwargs):
 
         response = {"action_id": instance.action.pk}
 
-        if settings.ACTIONNETWORK_SYNC:
-            sync_reminderrequest_to_actionnetwork.delay(instance.uuid)
+        reminderrequest_followup.delay(instance.uuid)
```
nickcatal, line 31: Instead of `instance.uuid` can you use `instance.pk`

It's the same thing, just more PK-agnostic

app/absentee/tasks.py
liewegas: This task was deprecated a week or two ago, so it shouldn't appear in the celery queue, right?  IIUC the old task has to stick around for just one red/black deployment cycle (or two in case we roll back more than 1 release)

nickcatal: Oh perfect, then this is good.

PR #382, opened 2020-08-14
geocode: behave if address=None is passed

liewegas:


liewegas:
Fixes failure when API endpoint hit with no params or data

PR #393, opened 2020-08-17
verifier: return last_updated even on an alloy negative result

liewegas:


app/verifier/alloy.py
```diff
@@ -73,3 +74,11 @@ def query_alloy(serializer_data):
...
 
     return response.json()
+
+
+def get_alloy_state_freshness(state):
+    response = requests.get(
+        ALLOY_FRESHNESS_ENDPOINT,
+        auth=requests.auth.HTTPBasicAuth(settings.ALLOY_KEY, settings.ALLOY_SECRET),
+    )
+    return response.json().get("data", {}).get("data_freshness", {}).get(state)
```
nickcatal, line 84: Can we both wrap this in an APM trace and add some sort of cache?

app/verifier/alloy.py
```diff
@@ -73,3 +75,16 @@ def query_alloy(serializer_data):
...
+
+
+def get_alloy_state_freshness(state):
+    freshness = cache.get("alloy_freshness", None)
+    if not freshness:
+        response = requests.get(
+            ALLOY_FRESHNESS_ENDPOINT,
+            auth=requests.auth.HTTPBasicAuth(settings.ALLOY_KEY, settings.ALLOY_SECRET),
+        )
+        freshness = response.json().get("data", {}).get("data_freshness", {})
```
nickcatal, line 87: Can you wrap this in APM for the alloy service

PR #415, opened 2020-08-25
actionnetwork: sync mobile numbers

liewegas:
first_party -> always 'subscribed'
partner -> respect sms_opt_in_partner


app/integration/actionnetwork.py
```diff
@@ -219,7 +219,6 @@ def _sync_item(item, subscriber_id):
     )
     info = {
         "person": {
-            "identifiers": [f"voteamerica_action:{item.action_id}"],
```
nickcatal, line 222: Can you remove any mention of the identifier from the action network integration documentation? And mention that we pass phone numbers?

app/integration/actionnetwork.py
```diff
@@ -232,6 +231,10 @@ def _sync_item(item, subscriber_id):
                     "country": "US",
                 },
             ],
+            "phone_numbers": [{
+                "number": str(item.phone),
+                "status": "subscribed" if not subscriber_id or item.sms_opt_in_partner else "unsubscribed",
```
nickcatal, line 236: Actually, will this unsubscribe a user who is subscribed to a subscriber?

liewegas: If they are subscribed, and they do not check the sms box, then yes, this will unsubscribed them from the subscriber's mobile list.  Let me confirm that passing None here works as intended.

PR #416, opened 2020-08-25
Support Jenkins in remote_run.sh

benweissmann:

PR #417, opened 2020-08-26
V2 ECS task defs

benweissmann:
Because we read the ECS task defs from `master`, I wanted to leave the existing ones in place until we've migrated over to the new ones with separate tasks per-service.

This adds a new `v2` directory for our new ECS tasks. The changes are:
- No more migrations! Migrations are now handled by Spinnaker during the deploy automatically
- Simplified the secrets to just have one set of secrets because we are breaking things out and shouldn't need to optimize the number of secrets on the Beat container anymore
- Separate ECS tasks for beat, web, and worker (each with their own main, datadog, and logging containers)

To reduce merge conflicts, this also includes the remote_run.sh updates from https://github.com/vote/turnout/pull/416


benweissmann:
Closing in favor of https://github.com/vote/turnout/pull/417

PR #425, opened 2020-08-28
Document translation of fields to 2020 version

amyrlam:
Story: https://app.clubhouse.io/turnout2020/story/3024/document-legacy-fields-2020-fields
Link to Credit Karma Slack convo: https://voteamerica.slack.com/archives/C017LH2CVHN/p1598475144072700?thread_ts=1598473785.068700&cid=C017LH2CVHN

***

* Note: I'm not sure how to run this locally, checked the README but figured I'd just make a PR to get the ball rolling. 
* Not sure about the name "legacy field" cause the non-2020 field is not being retired AFAIK (although it sounds like future fields will be `2021_`.
* Do we want to consider naming the fields with the year as the suffix down the road? e.g. `early_voting_ends_2021`?


amyrlam:
I think this probably is not showing up in the LHS bar, but I can check it out locally tomorrow.

PR #434, opened 2020-08-28
allow turnout to deploy proxies with uptime monitoring off

liewegas:
- new optimizely flag for checking sites
- proxy deployment dependent on api key instead


liewegas:
easier to include this in the voter pr

PR #445, opened 2020-09-01
Fix capitalization of field in Lob emails

amyrlam:
Will be written as "Postmarked by October 11, 2020" or "postmarked by October 11, 2020" in the Admin Tool. Having the "by" be capitalized seems fine, if that's what capfirst does?

Follow-on to: https://github.com/vote/turnout/pull/441 and https://github.com/vote/turnout/pull/442

Story: https://app.clubhouse.io/turnout2020/story/2518/front-end-designs-mockups-for-print-and-mail-api

WAS: (same in Lob email and other register emails)
image

***

I'm pretty rusty on Django, I'm not sure how to test this locally. I'm thinking it will return `Postmarked by October 11, 2020` or `Postmarked By October 11, 2020` 

What I really wanted to do was lowercase the first character of the strong only - since we aren't sure if the admin tool will be entered as `Postmarked by October 11, 2020` or `postmarked by October 11, 2020`. But added a colon here as a "quick fix".

### DIFF
--- a/app/register/templates/register/email/file_notification.html
+++ b/app/register/templates/register/email/file_notification.html
@@ -44,7 +44,7 @@ Just a few more steps to complete your voter registration.
 
 <hr><br>
 
-Your voter registration form must be <strong>{{ state_info.2020_registration_deadline_by_mail|markdownify|lower }}</strong>.<br><br>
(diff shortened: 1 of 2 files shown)


liewegas:
They just changed these to lowercase.

PR #448, opened 2020-09-01
send transactional emails from VoteAmerica, or Subscriber via VoteAmerica

liewegas:


app/multi_tenant/models.py
```diff
@@ -78,6 +78,13 @@ def full_email_address(self) -> str:
         clean_name = self.name.replace('"', "'")
         return f'"{clean_name}" <{self.email}>'
 
+    @property
+    def transactional_from_email_address(self) -> str:
+        if self.is_first_party:
+            return self.full_email_address
+        clean_name = self.name.replace('"', "'")
+        return f'"{clean_name} via VoteAmerica" <hello@voteamerica.com>'
```
liewegas, line 86: I don't really want to hard-code this, but it doesn't seem appropriate to load up the VA subscriber record.

PR #457, opened 2020-09-02
integration/mymove: pull leads every 3 hours; sync to actionnetwork manually

liewegas:


app/integration/mymove.py
```diff
@@ -0,0 +1,227 @@
...
+    for lead in leads:
+        # avoid duplicates
+        created_at = datetime.datetime.strptime(
+            lead["created_at"], "%Y-%m-%dT%H:%M:%SZ"
+        ).replace(tzinfo=datetime.timezone.utc)
+        if MymoveLead.objects.filter(
+            email=lead.get("email"), mymove_created_at=created_at
+        ).exists():
+            dup += 1
+            continue
```
nickcatal, line 74: 2 options here that may streamline things:

First this will keep your exists and create query to 1 line
```python
lead, created = MymoveLead.objects.get_or_create(
    # items to match on
    email=lead.get("email"), mymove_created_at=created_at,
    # if items don't match, use these defaults to create it
    defaults={"first_name": lead.get("first_name"), "last_name": "...", "fieldname": "value"})
if not created:
  dup += 1
```

Alternatively, you can seriously cut down on the number of queries in general by doing a `MymoveLead.objects.bulk_create(objects_list, ignore_conflics=True)` like https://github.com/vote/turnout/blob/master/app/official/usvf.py#L93

Probably want to go the 2nd route, but it's harder to get a log output of what exactly your query did

app/integration/mymove.py
```diff
@@ -0,0 +1,227 @@
...
+                "mymove_move_date": lead.move_date.isoformat(),
+                "mymove_cross_state": lead.old_state_id != lead.new_state_id,
+            },
+        },
+        "action_network:referrer_data": {"source": "mymove",},
+    }
+    url = ACTIONNETWORK_ADD_ENDPOINT.format(form_id=form_id)
+    response = requests.post(
+        url, json=info, headers={"OSDI-API-Token": settings.ACTIONNETWORK_KEY,},
+    )
```
nickcatal, line 201: Can you wrap this inside a datadog tracer

nickcatal (commented):
Just a comment/change on how the bulk creation is done

nickcatal (approved):

PR #458, opened 2020-09-03
Fix lint; add travis test to ensure things are linted

liewegas:
- A pre-commit hook is slow
- Also, it is tricky to distinguish between the staged content (being committed) vs other changes in the working directory that aren't staged for commit
- A pre-push hook as the same problems, compounded by e.g. ``git push origin other_branch_im_not_on_now``

A CI test for the PR seems like a good balance.


nickcatal (approved):
Thanks for this

PR #476, opened 2020-09-07
Add high-priority task queue

benweissmann:
- Create a high-priority task queue for emails
- Route high-priority tasks to that queue
- Run both a high-priority worker and a regular worker in the worker task (we could create a separate task, but this is simpler because we can just scale both workers together)
- Increase the default worker concurrency to 8. We had it as 3 because the PDF generation jobs would peg our CPU, and we allocate 4 CPUs per task, so limiting to 3 concurrent tasks meant that we'd always have one CPU available for serving web requests. Because our tasks are no longer CPU-bound, and we don't run web and worker on the same tasks, we can now safely increase this value to run more tasks concurrently.


nickcatal (approved):
Nice!

PR #490, opened 2020-09-11
Use 2020 fields in email for more better UX 

amyrlam:
Ref: https://docs.voteamerica.com/api/2020/

Those fields are lowercased now: https://admin.voteamerica.com/manage/election/fieldinformationtype/2020_vbm_request_deadline_by_mail/. We don't want to apply `lower` helper or it will lowercase the "O" in "October".

//

The reason why I'm making the PR is from https://voteamerica.slack.com/archives/CQE2AH65A/p1599784206076600. I'm not sure if there should be an if block in the email from TX...and if it's even possible to check for `if state = TX` in the email, or if that type of conditional check should come before.

- [ ] TODO: reply to daniela


liewegas (approved):

amyrlam:
@liewegas thanks for the review, i sort of left it in draft because not sure what to do about the TX thing... but seems like that can be addressed separately. i will merge it in

PR #492, opened 2020-09-11
Allow re-submission of registrations

benweissmann:
Removes the queryset filters from both register and absentee


app/absentee/api_views.py
```diff
@@ -95,7 +95,7 @@ class BallotRequestViewSet(IncompleteActionViewSet):
     tool = "absentee"
     model = BallotRequest
     serializer_class = BallotRequestSerializer
-    queryset = BallotRequest.objects.filter(status=TurnoutActionStatus.INCOMPLETE)
+    queryset = BallotRequest.objects
```
nickcatal, line 98: You'll need `BallotRequest.objects.all()` and `Registration.objects.all()`

nickcatal (approved):

PR #499, opened 2020-09-11
Disable all NC VBM form filling

benweissmann:


michaelsnook:
This looks good. I will test it locally tonight.

michaelsnook (changes requested):
When I test this locally, it asks me for my municipality and then there are no options there...
Screen Shot 2020-09-12 at 12 41 28 PM

If there's a way to just SKIP the county select and just email me my form... that would probably be better, yes?

michaelsnook:
Confirmed on staging this works fine.

PR #501, opened 2020-09-11
register: lob cover pages for MI MT MS PA FL OH

liewegas:
These are the top 6 self-print states after TX


michaelsnook (dismissed):
Florida box 8: should say "but not required" not "by"

michaelsnook:
I know we use this language elsewhere but I wonder if we should avoid the use of the words "pre-filled" because legally we're not _pre_ filling (based on info we had previously) but directly filling from user input. e.g.

> *Complete the form*
> The information you entered on our website has been filled in. We've also included a blank form in case you need it.

michaelsnook (commented):
I'm not sure if this matters... but our data for MT says enter either your DL or last-4, which is unlike other states that only want the last-4 iff you have no DL... our API data says:

> Enter your Montana driver’s license number or the last 4 digits of your social security number. If you don't have either number, please write “NONE” in this space.

michaelsnook:
not sure if I'm using the "review" feature correctly, but I'm done reviewing all 6 files; Sage has already made the correction to FL, and the other two comments are about MT wording and the use of the word "pre-fill". None of these would block a launch IMO but I think we should change both.

michaelsnook (approved):

PR #504, opened 2020-09-12
Surface warnings, clean up copy around deadlines on PDF cover sheets

benweissmann:
- Surface `warnings_registration` and `vbm_warnings` in the transactional emails we send with forms
- Surface these warnings on the coversheets
- Clean up the presentation of the mailing deadlines on the coversheets

https://app.clubhouse.io/turnout2020/story/1244/surface-vbm-warnings-and-registration-warnings-on-the-coversheets-and-in-the-emails


app/absentee/generateform.py
```diff
@@ -268,7 +272,7 @@ def prepare_formdata(
         )
     except StateInformation.DoesNotExist:
         state_mail_deadline_2020 = "mailed as soon as possible."
-    form_data["2020_state_deadline"] = state_mail_deadline_2020
+    form_data["2020_state_deadline"] = f"Your form must be {state_mail_deadline_2020}"
```
liewegas, line 275: The print-and-forward covers already have the "Your form must be".. we need to update that as well

benweissmann:
@liewegas good catch -- I've just broken out the full-sentence version as a new field so the PDFs can use whichever one makes sense (and left the existing field as-is)

liewegas (approved):
Perfect!

PR #512, opened 2020-09-14
polling-place backend

liewegas:
- Not sure making this an ``ActionModel`` adds much since we aren't logging any events
- I made ``unstructured_address`` required since we'll always have it, although strictly speaking I don't think we need to require it.


app/polling_place/api_views.py
```diff
@@ -0,0 +1,44 @@
...
+        serializer.validated_data["address1"] = addr.get("address_line_1")
+        serializer.validated_data["address2"] = addr.get("address_line_2")
+        serializer.validated_data["city"] = addr.get("city")
+        serializer.validated_data["zipcode"] = addr.get("zip")
+        try:
+            serializer.validated_data["state"] = State.objects.get(
+                code=addr.get("state")
+            )
+        except State.DoesNotExist:
+            raise serializers.ValidationError({"lookup_result": "Invalid state"})
```
liewegas, line 38: Note that reminders doesn't do this (error nicely if the state doesn't exist).  Should we fix that too?

app/polling_place/serializers.py
```diff
@@ -0,0 +1,35 @@
...
+        minimum_necessary_fields = [
+            "lookup_result",
+            "unstructured_address",
+        ]
+        optional_fields = [
+            "address1",
+            "address2",
+            "city",
+            "state",
+            "zipcode",
```
nickcatal, line 24: You don't need to list these fields as they won't be POST-able to the serializer

app/polling_place/tests/test_api_views.py
```diff
@@ -0,0 +1,156 @@
...
+                "requires_request": True,
+                "url": {
+                    "en": "iwillvote.com/votebymail?state=PA",
+                    "es": "voyavotar.com/votebymail?state=PA",
+                },
+            },
+            "precinct_code": "Philadelphia 08-04",
+            "precinct_id": "376611",
+            "state": "PA",
+            "status": "success",
```
nickcatal, line 111: Can we store the `status` result inside the model?

app/polling_place/models.py
```diff
@@ -0,0 +1,26 @@
...
+from multi_tenant.mixins_models import SubscriberModel
+
+
+class PollingPlaceLookup(
+    ActionModel, SubscriberModel, TrackingModel, UUIDModel, TimestampModel,
+):
+    unstructured_address = models.TextField(null=True)
+    lookup_result = JSONField(null=True)
+
+    state = models.ForeignKey("election.State", on_delete=models.PROTECT)
```
nickcatal, line 16: We need to store situations where someone enters garbage and nobody can ever figure out the state that was addressed. So this should be nullable.

app/polling_place/models.py
```diff
@@ -0,0 +1,26 @@
...
+from common.utils.models import TimestampModel, TrackingModel, UUIDModel
+from common.validators import zip_validator
+from multi_tenant.mixins_models import SubscriberModel
+
+
+class PollingPlaceLookup(
+    ActionModel, SubscriberModel, TrackingModel, UUIDModel, TimestampModel,
+):
+    unstructured_address = models.TextField(null=True)
+    lookup_result = JSONField(null=True)
```
nickcatal, line 14: We may have both a DNC response and a Google Civic response. maybe we store them as separate fields, both being able to be blank?

app/polling_place/api_views.py
nickcatal: Yeah, we should. Although in this case we're looking to just dump the data into the JSONField so I don't think we should be validating state at all.

app/polling_place/tests/test_api_views.py
```diff
@@ -0,0 +1,156 @@
...
+from multi_tenant.models import Client
+from polling_place.models import PollingPlaceLookup
+
+PP_API_ENDPOINT = "/v1/polling-place/lookup/"
+VALID_LOOKUP = {
+    "unstructured_address": "1234 spruce st., philadelphia, PA 19103",
+    "embed_url": "https://www.greatvoter.com/location/of/embed",
+    "session_id": "7293d330-3216-439b-aa1a-449c7c458ebe",
+    "lookup_result": {
+        "data": {
```
nickcatal, line 17: I actually encapsulate the result inside the `data` property inside of lambda. The core DNC response is everything inside of `data`, so we likely want to just pass to the backend whatever is inside `data`, on the off chance that we ever directly integrate turnout with the DNC.

nickcatal (commented):
Few notes

app/turnout/api_urls.py
```diff
@@ -16,4 +16,5 @@
     path("apikey/", include("apikey.api_urls")),
     path("leouptime/", include("leouptime.api_urls")),
     path("integration/", include("integration.api_urls")),
+    path("polling-place/", include("polling_place.api_urls")),
```
nickcatal, line 19: Can you remove the dash here (so it's `/v1/pollingplace/`)

nickcatal (approved):
One flag then looks good to me

PR #522, opened 2020-09-15
lob: fix mail callback task names

liewegas:
It's e.g. register.tasks.send_print_and_forward_mailed


nickcatal (approved):
ohhh!!! this makes sense

PR #530, opened 2020-09-17
mymove: geocode (new) leads

liewegas:
We will need to geocode for the direct mail LEO.  I think we may as well geocode all leads.


liewegas:
included in #536

PR #534, opened 2020-09-17
Switch to rabbitmq from redis as the celery broker

nickcatal:
We were running into a bunch of random errors from using redis as our broker, especially after adding a separate queue. By moving to a more dedicated message broker such as rabbitmq we may find scaling to more queues more reliable.

Or it may be like lighting money on fire and solve absolutely nothing.


app/turnout/celery_beat.py
```diff
@@ -9,30 +9,3 @@
...
-def report_stats():
-    threading.Timer(settings.BEAT_STATS_INTERVAL, report_stats).start()
-
-    conn = redis.from_url(settings.CELERY_BROKER_URL)
-    queue_len = sum(conn.llen(name) for name in queue_names)
-
-    print(f"Queue Length: {queue_len}", flush=True)
-
-    if settings.BEAT_STATS_METRIC_NAMESPACE:
-        cloudwatch_client.put_metric_data(
```
benweissmann, line 25: Are we getting this into Cloudwatch somehow for autoscaling?

benweissmann (approved):

app/turnout/celery_beat.py
nickcatal: Yeah, via cloudamqp

PR #537, opened 2020-09-18
voter: fix up rechecks

liewegas:
- extra args so we can trigger these easily by state etc
- timebox tasks
- requery based on the alloy data updates
- post to slack when alloy data updates


app/voter/tasks.py
```diff
@@ -256,15 +257,20 @@ def lookup(item):
 
 
 @shared_task
-def check_new_actions(limit=None, state=None):
+def check_new_actions(
+    limit: int = None, state: str = None, max_minutes: int = None
+) -> None:
```
nickcatal, line 262: How is this going to handle instances where, due to the red/black deployment, there are task requests in-flight during deployment that may not have the extra arg?

nickcatal (commented):
One Q

app/voter/tasks.py
liewegas: The new code has the defaults, so callers from old code should be fine.

The new code doesn't set max_minutes, so new callers hitting old code should be fine.

nickcatal (approved):

PR #548, opened 2020-09-22
official: behave if geocodio result doesn't include a state

liewegas:
This can happen for a very bad match, triggered by e.g. a bad zip code.


nickcatal (approved):
Weird situation. Thanks for handling this

PR #549, opened 2020-09-22
use url generate for all email URLs to voteamerica.com

liewegas:


app/register/templates/register/email/print_and_forward_returned.html
```diff
@@ -15,7 +15,7 @@
 
 <br>
 
-We're very sorry the letter was not delivered.  If the address above is incorrect, please <a href="https://www.voteamerica.com/register-to-vote/?{{query_params}}&refcode=register_print_and_forward_returned" target="_blank">use our registration tool again</a> with the correct mailing address for the registration form.
+We're very sorry the letter was not delivered.  If the address above is incorrect, please <a href="https://www.voteamerica.com/register-to-vote/?{{query_params}}&utm_medium=email&utm_source=turnout&utm_campaign=register-print-and-forward-returned&source=va_email_turnout_register-print-and-forward-returned&refcode=va_email_turnout_register-print-and-forward-returned" target="_blank">use our registration tool again</a> with the correct mailing address for the registration form.
```
nickcatal, line 18: I think the `{{query_params}}` will result in duplicate `utm_` codes.

nickcatal (approved):

app/register/templates/register/email/print_and_forward_returned.html
liewegas: Okay this comes up in a few places:
- the register -> absentee upsell email.  
- the verifier upsell texts for register and absentee
- the email when your letter was returned by USPS, linking you back to the tool to try again

In all of these cases, do we want the tracking to reflect the original tracking info, or should we use new tracking info saying it comes from the upsell email or text or notification or whatever?

nickcatal:
One flag, every time we link anywhere on voteamerica.com we can add the `subscriber` query param, and the subscriber will get credit for any transaction that happens during that visit. We should probably pass that along, even if we're just linking to the LEO tool.

ericisaiah:

PR #558, opened 2020-09-24
lob: process letter status webhook asynchronously

liewegas:


app/integration/api_views.py
```diff
@@ -43,69 +41,11 @@ def wrapper(request, *args, **kwargs):
...
-                if days:
-                    DelayedTask.schedule_days_later_polite(
-                        item.state.code, days, task, str(item.pk), str(action.pk)
-                    )
-                else:
-                    DelayedTask.schedule_polite(
-                        item.state.code, task, str(item.pk), str(action.pk)
-                    )
-
+    process_lob_letter_status.delay(letter_id, etype)
```
nickcatal, line 50: Can you add some sort of random countdown here? This task hits the database a few times and there is no reason to spike up our database/task load if Lob decides to dump a huge number of notifications on us. Might not be the worst idea to add some randomness to the `DelayedTask` as well (we could skew those over 60 minutes, for example.)

nickcatal (commented):
One comment about timing

nickcatal (approved):

PR #559, opened 2020-09-24
register lob covers: IA, NV, VA, ME

liewegas:


michaelsnook (changes requested):
In NV we could save one vertical line by changing wording to:

> For box 7 (Party choice): You must register with a political party ~if you want~ to take part in that party’s primary election, caucus, or convention. If you register as a nonpartisan, you will receive a nonpartisan ballot for the primary ~election~.

But I leave it to you whether to care about this detail.

For Virginia, this wording seems like a mistake?
>For box 6 (ID number): Virginia requires you to provide your full social security
number to register to vote. It will appear on reports produced only for official use
by voter registration and election officials and, for jury selection purposes, _the
courts_.

Maybe just cut to: "by voter registration and election officials and for jury selection."

You could also shave a line like this:
> Virginia requires ~you to provide~ your full social security number to register to vote.

Iowa, again, can shave a line like so:
> You may register with a party in advance ~if you want~ to take part in that party's primary election.

All the line-shaving suggestions are optional, only the one VA issue appears to be a mistake.

michaelsnook (approved):
Looks good to me!

PR #563, opened 2020-09-24
create a bunch of queues

liewegas:
- new worker class "bulk" that is intended to have all of the huge queues with high latency (and should probably autoscale differently)
- create a bunch of new queues
- annotate either entire subsystems (e.g., leouptime) or individual tasks (per-item work, but not the tasks that (will) queue them)


ecs/v2/template/service_worker.jsonnet
```diff
@@ -11,7 +11,11 @@ local memory = std.extVar('memory');
   containerDefinitions: [
     turnoutContainer.common('turnoutworker', 'celery', '/app/ops/worker_health.sh || exit 1') + {
       name: 'worker',
-      command: ['/app/ops/worker_launch.sh', 'default'],
+      command: ['/app/ops/worker_launch.sh', 'default,leouptime,usvf'],
+    },
+    turnoutContainer.common('turnoutworker', 'celery', '/app/ops/worker_health.sh || exit 1') + {
+      name: 'workerbulk',
+      command: ['/app/ops/worker_launch.sh', 'voter,movers,actionnetwork,lob-status-updates'],
```
nickcatal, line 18: You need to build the ecs task definitions after you add this

Then we need to add them to spinnaker.

nickcatal (commented):
Need to build the ECS definitions, then we may want to adjust the worker concurrency until we have a sense of memory usage with this 3rd container (or break this out to a 2nd type of worker/server group. Which is def a T4 thing.

ecs/v2/template/service_worker.jsonnet
liewegas: The bulk workers is definitely an optional thing.. not sure if it is needed now.  It was helpful locally to make sure that the queue routing was working correctly

nickcatal (approved):

PR #564, opened 2020-09-24
task fixes

liewegas:
- per-item tasks
- fix voter recheck query


app/integration/movers.py
```diff
@@ -450,9 +457,17 @@ def get_form_id(ids):
             an_id = get_form_id(form["identifiers"])
             if an_id:
                 logger.info(f"Found existing {form_name} form {an_id}")
+                cache.set(cache_key, an_id)
```
nickcatal, line 460: Do we want the value to persist for a long period of time? I believe by default the django cache will expire things promptly.

app/integration/movers.py
```diff
@@ -559,7 +581,7 @@ def push_to_actionnetwork(limit=None, offset=0, new_state=None) -> None:
         end = None
 
     for lead in q[offset:end]:
-        push_lead(session, form_id, lead)
+        _push_lead(session, form_id, lead)
         time.sleep(settings.ACTIONNETWORK_SYNC_DELAY)
```
nickcatal, line 585: Let's remove the sync delay after we split things out and add celery rate limiting.

app/integration/tasks.py
```diff
@@ -97,6 +97,22 @@ def geocode_movers(old_state: str = None, new_state: str = None):
...
+    lead = MoverLead.objects.get(pk=mover_pk)
+    geocode_lead(lead)
+
+
+@shared_task
+def push_mover_to_actionnetwork(mover_pk: str) -> None:
+    from .movers import push_lead
+
+    lead = MoverLead.objects.get(pk=mover_pk)
+    push_lead(lead)
```
nickcatal, line 113: Where do we call `push_mover_to_actionnetwork`?

nickcatal (commented):
Few Qs

app/integration/movers.py
liewegas: Sure, although it doesn't really matter--this change is just so we don't hit the endpoint on every iteration.

app/integration/movers.py
liewegas: Yeah--I think in the end this code will go away entirely, or we'll be putting items on the queue here

app/integration/tasks.py
liewegas: Nowhere yet.  This PR is just adding the single-item tasks so I can play with the code that is doing the queuing.

nickcatal: 👍

nickcatal (approved):

PR #565, opened 2020-09-24
lob: suggest 6-10 days

liewegas:


amyrlam:
@liewegas shall i change this on the frontend also? for register and absentee?

nickcatal (approved):

liewegas:
> @liewegas shall i change this on the frontend also? for register and absentee?

Yes!  I was going to do that too but then I had to make dinner :)

amyrlam:
@liewegas all good, i'll fix it on register and the absentee branch that i'm still working on today

PR #572, opened 2020-09-27
queue bulk work en masse

liewegas:
- When we have a zillion records to process, queue them as async items in their own queues
- (Re)queue items needing processing at given interval
- Make queued items expire a just under that same interval
- Optimizely variables to control how many items get queued
- Throttles on the work tasks.


app/integration/tasks.py
```diff
@@ -91,21 +71,21 @@ def push_movers_to_actionnetwork(limit=None, offset=0, new_state=None) -> None:
 
 
 @shared_task
-def geocode_movers(old_state: str = None, new_state: str = None):
+def geocode_movers(old_state: str = None, new_state: str = None, limit: int = None):
     from .movers import geocode_leads
 
-    geocode_leads(old_state=old_state, new_state=new_state)
+    geocode_leads(old_state=old_state, new_state=new_state, limit=limit)
```
nickcatal, line 77: Is this going to make things go nuts during the red/black deployment?

app/integration/movers.py
```diff
@@ -177,9 +180,16 @@ def send_blank_register_forms_tx(offset=0, limit=None) -> None:
     end = None
     if limit:
         end = offset + limit
-    print(f"offset {offset} limit {limit} count {q.count()}")
+    queue_async = get_feature_bool("movers", "blank_forms_tx_async")
+    max_minutes = get_feature_int("movers", "blank_forms_tx_max_minutes") or 55
+    logger.info(
+        f"send_blank_register_forms_tx offset={offset}, limit={limit}, count={q.count()}, async={queue_async}, max_minutes={max_minutes}"
```
nickcatal, line 186: lint?

app/voter/tasks.py
```diff
@@ -406,13 +417,19 @@ def _recheck_old_actions(
...
-            logger.info(f"Hit max runtime ({max_minutes} minutes)")
-            break
+        else:
+            item = action.get_source_item()
+            if item:
+                lookup(item)
+                checked += 1
+            if stop < datetime.datetime.utcnow():
+                logger.info(f"Hit max runtime ({max_minutes} minutes)")
+                break
```
nickcatal, line 432: Why not just execute `voter_lookup_action` directly, without the apply_async? Then it's a consistent experience if async or not

nickcatal (commented):
Few Qs

app/integration/tasks.py
liewegas: in this case no--most of these are still off in optimizely

app/voter/tasks.py
liewegas: One of the goals here is to parallelize the lookups (against alloy mostly, but also TS).

app/integration/movers.py
liewegas: that was linted.. i'll break up the long line tho!

nickcatal (approved):

PR #575, opened 2020-09-28
voter: create voter record for misc non-"Active" alloy states

liewegas:
As long as we have a alloy_person_id and the status is
in the acceptable list, do the match.

Do NOT match the (very common) 'ZIP Not Yet Supported'
(not sure what that means!).


nickcatal (approved):

nickcatal:
"Zip Code Not Supported" is a CA thing

PR #591, opened 2020-09-30
Update AK PDF

benweissmann:
https://app.clubhouse.io/turnout2020/story/3829/alaska-drops-part-of-the-mailing-address-for-print-and-mail


amyrlam (approved):
thx!

PR #615, opened 2020-10-08
locator export/report

liewegas:
Main caveat here is the `county` and `precinct_id` columns are only populated if there is a dnc_result--civic result doesn't include that info.


app/reporting/migrations/__init__.py
```diff
@@ -149,10 +149,60 @@
...
+    lookup.civic_status,
+    CASE WHEN lookup.dnc_result IS NOT NULL THEN
+      lookup.dnc_result->'data'->'state'
+    ELSE
+      lookup.civic_result->'normalizedInput'->'state'
+    END as state_id,
+    CASE WHEN lookup.dnc_result IS NOT NULL THEN
+      lookup.dnc_result->'data'->'county'
+    ELSE
+      NULL
```
nickcatal, line 172: What happens here if there is junk passed to the dnc_result field? We don't do a ton of validation on this API endpoint.

nickcatal (approved):
One Q. If it's not an issue good by me.

app/reporting/migrations/__init__.py
liewegas: if the JSON field isn't present then it'll be NULL in the view

PR #631, opened 2020-10-20
smsbot: simple SMS blast capability; blast polling place map via MMS

liewegas:
image


app/smsbot/tasks.py
```diff
@@ -61,3 +64,22 @@ def poll_twilio() -> None:
...
         poll()
+
+
+@shared_task(queue="blast-sms")
+def send_blast_sms(number_id: str, blast_id: str) -> None:
+    from .blast import send_blast_sms
+
+    number = Number.objects.get(pk=number_id)
+    blast = Blast.objects.get(pk=blast_id)
+    send_blast_sms(number, blast)
```
nickcatal, line 75: Can we call this task something different than what we import?

app/turnout/settings/core.py
```diff
@@ -248,6 +248,8 @@
     "movers": 1 / 5,
     "geocode": 1 / 10,
     "lob-status-updates": 1 / 10,
+    "blast-mms": 1 / 30,
+    "blast-sms": 1 / 100,
```
nickcatal, line 252: We can go higher on SMS-per-second. And should probably limit to like 28 MMS-per-second or something maybe, incase someone else tries to send a MMS at the same time Turnout is churning along?

app/smsbot/blast.py
```diff
@@ -0,0 +1,102 @@
...
+        f"{DNC_API_ENDPOINT}?{urlencode({'address': home_address})}"
+    )
+    pp = response.json().get("data", {}).get("election_day_locations", [])
+    if not pp:
+        logger.info(f"No election day location for {voter}")
+        return
+    pp = pp[0]
+    pp_address = f"{pp['address_line_1']}, {pp['city']}, {pp['state']} {pp['zip']}"
+    pp_loc = f"{pp['lon']},{pp['lat']}"
+    pp["location_name"]
```
nickcatal, line 57: Does this do anything?

app/smsbot/blast.py
```diff
@@ -0,0 +1,102 @@
...
+
+    if HIRES:
+        size = "512x512"
+    else:
+        size = "256x256"
+        zoom -= 1.0
+
+    centerx = (float(pp["lon"]) + float(home[0]["location"]["lng"])) / 2
+    centery = (float(pp["lat"]) + float(home[0]["location"]["lat"])) / 2
+    map_loc = f"{centerx},{centery},{zoom}"
```
nickcatal, line 81: ^ is all great

app/smsbot/blast.py
```diff
@@ -0,0 +1,102 @@
...
+If you are still registered at
+ {home_address.upper()}
+then your polling place is:
+
+{pp['location_name'].upper()}
+{pp_address.upper()}
+{pp['dates_hours']}
+
+        {shorten_url(locator_url)}
+""",
```
nickcatal, line 98: Let's chat through final language before we send it, but for now this is good

app/smsbot/blast.py
```diff
@@ -0,0 +1,102 @@
...
+    if not home:
+        logger.info(f"Failed to geocode {voter}")
+        return
+    home_address = f"{voter.address_full}, {voter.city}, {voter.state} {voter.zipcode}"
+    home_loc = f"{home[0]['location']['lng']},{home[0]['location']['lat']}"
+
+    # geocode polling place
+    response = requests.get(
+        f"{DNC_API_ENDPOINT}?{urlencode({'address': home_address})}"
+    )
```
nickcatal, line 49: There needs to be an `x-api-key` header, and we also need to pass along the query param `client_id` equal to `client_id_20`

Also `requests.get()` has built in functionality to set query params

```
payload = {'key1': 'value1', 'key2': ['value2', 'value3']}
r = requests.get('https://httpbin.org/get', params=payload)
```

app/smsbot/models.py
```diff
@@ -98,3 +108,36 @@ def send_sms(self, text, ignore_opt_out=False):
...
+        return f"Blast - {self.description} ({self.uuid})"
+
+    def enqueue_sms_blast(self):
+        from .tasks import send_blast_sms
+
+        logger.info(f"Enqueueing {self}")
+        num = 0
+        for n in Number.objects.raw(self.sql):
+            send_blast_sms.delay(str(n.pk), self.pk)
+            num += 1
```
nickcatal, line 131: This could be slow. Do we have any way to recover if this code fails early that won't result in the user getting another MMS?

nickcatal (commented):
Few comments

app/turnout/settings/core.py
liewegas: How high?

app/smsbot/models.py
liewegas: 1- The final task refuses to send a dup message within the same blast
2- I think we can fiddle with the sql statement to exclude records that have already been sent

app/turnout/settings/core.py
nickcatal: 300/second? We have a 500/second cap and ActionNetwork can go as fast as 150/second.

nickcatal (approved):

PR #634, opened 2020-10-22
usvf: do not use bulk_update()

liewegas:
This fails without batch_size on aurora. And even if it didn't, it
doesn't update modified_at, which is very handy. Open code our own
version.

Also.. this version seems to be faster??

Note that adding batch_size=100 to the bulk_update() would probably also work, but that doesn't get us the modified_at update.


nickcatal (approved):
weird... didn't expect the db move to cause this.

liewegas:
> weird... didn't expect the db move to cause this.

It's just a guess, but seems like the most plausible difference between local postgres and production.

PR #636, opened 2020-10-24
official: expose lat/lng in address serializer

liewegas:


app/official/serializers.py
```diff
@@ -10,6 +10,19 @@ class Meta:
 
 
 class AddressSerializer(serializers.ModelSerializer):
+    lat = serializers.SerializerMethodField("get_lat")
+    lng = serializers.SerializerMethodField("get_lng")
```
nickcatal, line 14: Would this allow people to POST the fields `lat`/`lng`, which will be passed into the `get_lat`/`get_lng` methods as strings/integers, which wouldn't have a `.location`?

liewegas: it's read-only, see first sentence of https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield

ericisaiah:

PR #660, opened 2020-11-06
Move subscriber stats into postgres and recompute incrementally

benweissmann:


app/reporting/tasks.py
```diff
@@ -39,90 +41,74 @@ def send_report_notification(report_pk: str):
...
-
-    cache.set(f"client.stats/{uuid}", r, 24 * 60 * 60)
+    with transaction.atomic():
+        with connection.cursor() as cursor:
+            # We fetch stats starting from the last update time, and going until (now - 1 minute).
+            # The 1 minute offset is to account for differences between the DB clock and
+            # the app server clock -- if the app server is ahead of the DB, we don't
+            # want to think that we have data up to, e.g. 4PM if the DB thinks it's 3:59PM.
+            current_update_time = datetime.now(tz=timezone.utc) - timedelta(minutes=1)
+            refresh_obj = StatsRefresh.objects.all()[0]
```
liewegas, line 51: s/all()[0]/first()/ ?

liewegas (approved):
love it!

nickcatal (approved):

app/reporting/tasks.py
benweissmann: This will error if there are no `StatesRefresh` rows, which I think is desirable (I think `.first()` returns None, which will lead to more confusing errors below rather than a clearer error here)

PR #669, opened 2020-11-10
Reword SMS opt-in options and fix some bugs

benweissmann:
- Rework SMS opt-in options to be more accurate about what's allowed under TCPA
- Default to an unchecked checkbox
- Rework the styling on those radios a bit
- Add `sms_mode` to the `fields` array of the interest form (without this, it looks like we have been ignoring that part of the signup form entirely and defaulting every new subscriber to an unchecked checkbox regardless of what they picked on the interest form)


app/subscription/templates/subscription/register.html
```diff
@@ -10,8 +10,21 @@
...
+                            {{ form.email|as_crispy_field }}
 
+                            <h3>Subscription</h3>
+                            {{ form.product|as_crispy_field }}
+                            {{ form.nonprofit|as_crispy_field }}
+                            {{ form.ein|as_crispy_field }}
+
+                            <h3>SMS Program</h3>
+                            {{ form.sms_mode|as_crispy_field }}
+                            <small class="form-text text-muted manual-help">**Please consult a TCPA attorney before selecting this option.</small>
```
liewegas, line 27: ``Please consult a TCPA attorney before adjusting this option``? Since the default is safe for anyone?

liewegas (approved):
LGTM

nickcatal (approved):

PR #672, opened 2020-11-11
optimizely: fix refresh for celery workers

liewegas:


nickcatal (approved):
Let's get this on staging and make sure that everything shows up in datadog as expected

PR #678, opened 2020-11-20
leouptime: add block string from OR

liewegas:


ericisaiah:

PR #681, opened 2020-12-01
Add /v1/people/opt-out/ endpoint for email and/or phone

liewegas:
Allow opt-out by email or by phone (or both).

No feedback is provided to the caller.  The AN update is done async in
the appropriate worker queue.

### DIFF
--- a/app/integration/tasks.py
+++ b/app/integration/tasks.py
@@ -66,10 +66,17 @@ def sync_subscriber_to_actionnetwork(pk: str) -> None:
 
 
 @shared_task(queue="actionnetwork")
-def unsubscribe_phone_from_actionnetwork(phone: str) -> None:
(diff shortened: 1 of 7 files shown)


app/integration/tasks.py
```diff
@@ -65,6 +65,13 @@ def sync_subscriber_to_actionnetwork(pk: str) -> None:
     sync_subscriber(subscriber)
 
 
+@shared_task(queue="actionnetwork")
+def unsubscribe_email_from_actionnetwork(email: str) -> None:
```
benweissmann, line 69: We should add retry to this, like the email tasks (so we can handle AN going down for a bit)

liewegas: The session in actionnetwork.py already has a retry, with a backoff of a few seconds... do you think we also should add a retry (in minutes) on the task?

PR #683, opened 2020-12-08
Implement new subscription plans

benweissmann:
This PR:

- replaces the old DB-configured subscription terms with a new set of three plans: free, premium, and nonprofit.
- blocks users on the free plan from exporting their data or seeing voter-level results
- adds contact information to the subscription itself so we know who to contact about the subscription


app/integration/actionnetwork.py
```diff
@@ -368,24 +368,22 @@ def sync_subscriber(subscriber: Client) -> None:
...
         "person": {
-            "given_name": interest.first_name,
-            "family_name": interest.last_name,
-            "email_addresses": [{"address": interest.email}],
+            "given_name": subscription.primary_contact_first_name,
+            "family_name": subscription.primary_contact_last_name,
+            "email_addresses": [{"address": subscription.primary_contact_email}],
             "custom_fields": {
-                "subscriber_is_c3": interest.nonprofit,
+                "subscriber_is_c3": subscription.plan == enums.SubscriberPlan.NONPROFIT,
```
benweissmann, line 383: @liewegas how can i test this? And do i need to pre-define these custom fields in AN?

Also, maybe we want to have this sync run nightly to pick up changes?

liewegas: There is a `Dev subscribers` actionnetwork group with an API key you can put in `.env` as `ACTIONNETWORK_SUBSCRIBERS_KEY` (along with `ACTIONNETWORK_KEY`).  You may also need to set `ACTIONNETWORK_SYNC=True`

docker-compose.yaml
```diff
@@ -196,15 +197,15 @@ services:
...
-      - "9003:8000"
+  # docs:
+  #   build:
+  #     context: .
+  #     dockerfile: Dockerfile-docs
+  #   command: mkdocs serve -a 0.0.0.0:8000
+  #   volumes:
+  #     - ./docs:/docs
+  #   ports:
+  #     - "9003:8000"
```
liewegas, line 208: Did you mean to include the docker-compose changes?

benweissmann: Yes -- forgot to mention these changes in the PR description. I was having a lot of issues with my dev environment (very slow to reload), and these changes helped somewhat -- i just commented out non-critical containers (so you can un-comment them if you need them but they don't run by default) and combined all the workers into a single worker. Does that seem OK to you?

liewegas: Yep those changes look good!

app/subscription/forms.py
```diff
@@ -1,18 +1,32 @@
...
+                enums.SubscriberPlan.FREE,
+                "Free: Embed VoteAmerica's tools on your own website and see aggregated stats, at no cost.",
+            ),
+            (
+                enums.SubscriberPlan.PREMIUM,
+                "Premium - $2,000/mo: Get detailed info on the voters who use VoteAmerica tools you embed on your website or link to, including email addresses and phone numbers to build your lists.",
+            ),
+            (
+                enums.SubscriberPlan.NONPROFIT,
+                "Nonprofit: 501(c)3 charitable organizations can use VoteAmerica's premium tools at a discount. Free until the end of 2020.",
```
liewegas, line 25: Do we still want this 'free until end of 2020' bit?

liewegas (approved):
LGTM

app/subscription/models.py
```diff
@@ -55,6 +56,10 @@ class Interest(
         null=True,
     )
 
+    # DEPRECATED
+    nonprofit = models.BooleanField(null=True)
+    product = models.ForeignKey(Product, null=True, on_delete=models.SET_NULL)
```
nickcatal, line 61: Let's make sure we remove these columns after this is deployed.

app/subscription/forms.py
```diff
@@ -1,18 +1,32 @@
...
-        label="Term",
-        queryset=Product.objects.filter(public=True),
+    sms_mode = sms_mode_choice
+
+    plan = forms.ChoiceField(
+        choices=[
+            (
+                enums.SubscriberPlan.FREE,
+                "Free: Embed VoteAmerica's tools on your own website and see aggregated stats, at no cost.",
+            ),
```
nickcatal, line 18: Can we remove this option on the interest form for right now? We need the subscription level available, but not sure that we want to explicitly offer it as an option when an account is setup right now.

nickcatal (approved):
Only flag is that we shouldn't offer an option for people to signup for a free tier account just yet. Otherwise, good!

PR #697, opened 2021-01-08
Move uptime monitoring to an external service

liewegas:
This removes the leouptime app, and instead configures an external uptime monitoring system.


Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY app/package.json app/package-lock.json $APP_DIR/
 RUN npm install
 
 COPY app/requirements.txt $APP_DIR/
-RUN pip install -r requirements.txt --use-deprecated=legacy-resolver
+RUN ( pip --version ; echo "pip 20.3 xxxx") | sort | head -n 1 | grep -q xxxx && pip install -r requirements.txt --use-deprecated=legacy-resolver || pip install -r requirements.txt
```
nickcatal, line 11: What does this do?

nickcatal (approved):
One question, then good

Dockerfile
liewegas: This option fails on my machine.  Not sure why since it should be platform independent but :shrug:

PR #698, opened 2021-01-14
Management commands for search/replace in the API

benweissmann:
Add two management command for doing global search and global find/replace across all API fields.


nickcatal (commented):
Can you have this use django-revision (inside the app) so that we have a record of the change?

benweissmann:
@nickcatal updated

nickcatal (approved):
Great!

PR #699, opened 2021-01-14
Remove alloy support

liewegas:
Leave columns in the various models/tables to preserve historic data.


nickcatal (approved):
Thanks for this

PR #709, opened 2021-02-09
Add individual date-of-birth fields

benweissmann:
Allows the client to pass separate year_of_birth, month_of_birth, and day_of_birth fields. Implements the validation logic from https://github.com/vote/voteamerica/blob/staging/src/utils/validations.js.


nickcatal (approved):
This is great!

PR #711, opened 2021-02-10
Zipcode and phone validation

benweissmann:
This implements the zipcode and phone number cleaning/validation logic from the frontend, as well as inferring gender from title (which we currently do entirely on the frontend, except for the Action API where we do it on the backend -- now we'll do it entirely on the backend): https://github.com/vote/voteamerica/blob/staging/src/utils/validations.js


app/common/utils/fields.py
```diff
@@ -18,3 +22,45 @@ def to_internal_value(self, data):
...
+        processed_phone = re.sub(r"\D", "", data)
+        if re_phone_digits_only.match(processed_phone):
+            phone_number = phone_to_python(f"+1{processed_phone}")
+        elif re_phone_digits_only_with_one.match(processed_phone):
+            phone_number = phone_to_python(f"+{processed_phone}")
+        else:
+            raise ValidationError("Enter a valid phone number2")
+
+        if phone_number and not phone_number.is_valid():
+            raise ValidationError("Enter a valid phone number3")
```
nickcatal, line 48: unnecessary 3 here

nickcatal (approved):
One typo

PR #717, opened 2021-02-23
Add authenticated info API

benweissmann:
- Add warnings to the election info API that the API is going away soon
- Create a new `/api/election/private/state/` endpoint with a warning that it's internal-only
- Create a new `/api/election/data/state` endpoint that uses API authentication, and Redis-based caching


app/election/api_views.py
```diff
@@ -22,6 +31,24 @@ def retrieve(self, request, pk=None):
...
+class SubscriberStateViewSet(StateViewSet):
+    serializer_class = SubscriberStateSerializer
+    authentication_classes = [ApiKeyAuthentication]
+    permission_classes = [ElectionApiAccessRequired]
+
+    @method_decorator(cache_page(60 * 15))
+    def list(self, request):
+        return super().list(request)
+
+    @method_decorator(cache_page(60 * 15))
```
nickcatal, line 47: Instead of using a cache decorator, I'd store the data inside redis directly then inside this signal receiver bust the cache https://github.com/vote/turnout/blob/master/app/election/models.py#L160-L163

That code will run any time the info model is changed, regardless of where it's updated (admin, API, etc)

nickcatal (commented):
One note about how we're doing caching

app/election/api_views.py
benweissmann: I'm a litle confused -- is there a problem with using the cache decorator? Or are you just saying that we should bust the cache when the data changes? Those two things seem orthogonal. I don't think there's an easy way to hook into the request/response cycle to cache the response JSON, other than using the cache decorator (or re-implementing whatever the cache decorator is doing inside my own middleware). Is there a problem with using the `cache_page` helper and then just calling `cache.delete` to bust the cache?

benweissmann: I'm also not convinced it's worth the complexity to deal with cache-busting; i just set a 15-minute cache expiration here and that's probably good enough?

nickcatal: So there are 2 concerns here

(1) The resource load that this page generates
(2) The content that is returned if it's cached
(3) The speed at which this endpoint returns data

Regarding the resource load, I'm not sure this cache tag will stop someone from entering /path/?1 /path/?2 /path/?3 and continually breaking the cache, putting load on the system. We could strip querystrings off with Cloudflare though, which would somewhat negate this problem. We just need to plan on it.

I'm also not sure that the decorator knows to cache based on the API key header, and won't start serving authenticated information to people with invalid API keys, or the warning page to people with valid API keys. Django's caching decorator isn't that smart by default (based on my past use of it.)

Regarding speed, I could absolutely see a world where API users are hitting the API less frequently than every 15 minutes, which would mean any requests are "cold" and could block rendering or parsing of partner API responses (I've seen instances where the main endpoint takes north of 1 second to return data, which would probably break the use case we saw where people were pulling data straight into Wordpress.)

benweissmann: re: auth -- from my testing, the authentication middleware comes before the caching middleware so auth is still applied even when the response is cached.

I'll go ahead and implement a custom cache middleware that uses a fixed key so we can ignore query params, and invalidate the cache manually -- it doesn't look like the built-in django cache middleware has a way of customizing which cache key to use.

nickcatal (approved):
good!

PR #718, opened 2021-03-01
Rework API key UI and API docs for the election API

benweissmann:
- Restrict access to the Action API (previously, we didn't need to do this because we only allowed subscribers with Action API access to create API keys)
- Open up API key management UI to all subscribers with either election API access (any subscriber not on the free plan) OR the action API
- Update the docs to reflect the new authenticated API endpoints


app/apikey/manage_views.py
```diff
@@ -23,7 +23,7 @@
 
 class ApiAccessViewMixin(SubscriberManageViewMixin):
     def check_subscriber_access(self):
-        if not self.subscriber.has_api_access:
+        if not self.subscriber.has_api_access():
             raise Http404
```
ericisaiah, line 27: I know this was not a change you made, but might be good to make these Http403 moving forward.

ericisaiah (commented):
Thanks for walking through it with me. Besides the minor comment about http codes, and addressing the CK issue, my only other question is about the OptOutView and OptInView classes. Should those also be protected by the action API authorization check?

benweissmann:
The OptInView and OptOutView are endpoint that we use internally; they have a check at the top for `request.auth.subscriber.is_first_party` (which indicates that the API key belongs to VoteAmerica, and not a subscriber) -- but we could refactor that check to be its own `FirstPartyApiKeyRequired` permission class.

ericisaiah:
Good idea - I'd vote for that to be consistent with auth across API endpoints.

benweissmann:
👍 updated

ericisaiah (commented):
Looks good

PR #719, opened 2021-03-03
Add futurevoter API endpoint

benweissmann:
- Adds a new endpoint for submitting FutureVoter records
- Refactors the DateSerializer a bit to handle more custom age validation, and use that to make sure FutureVoters are between 13 and 18
- Allow futurevoter.com in our CORS config


app/common/utils/serializers.py
```diff
@@ -104,4 +104,19 @@ def validate(self, data):
...
+                self.dob_validate(assembled_date)
+
+                data["date_of_birth"] = assembled_date
+        else:
+            self.dob_validate(data.get("date_of_birth"))
+
         return super().validate(data)
+
+    def dob_min_year(self):
+        return 1900
```
ericisaiah, line 116: Is there a reason this is a function vs a constant?

app/futurevoter/migrations/0001_initial.py
```diff
@@ -0,0 +1,46 @@
...
+
+    initial = True
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='FutureVoter',
+            fields=[
```
ericisaiah, line 18: I know it's not on futurevoter.com now, but I think we should try to collect Zip code, at least. Is there a reason we're not collecting any location data about the person? @debracleaver  There may be context I'm missing.

app/futurevoter/migrations/0001_initial.py
```diff
@@ -0,0 +1,46 @@
...
+                ('utm_medium', models.TextField(blank=True, null=True)),
+                ('utm_term', models.TextField(blank=True, null=True)),
+                ('utm_content', models.TextField(blank=True, null=True)),
+                ('session_id', models.UUIDField(blank=True, null=True)),
+                ('source', models.TextField(blank=True, null=True)),
+                ('email_referrer', models.TextField(blank=True, null=True)),
+                ('mobile_referrer', models.TextField(blank=True, null=True)),
+                ('first_name', models.TextField(null=True)),
+                ('last_name', models.TextField(null=True)),
+                ('date_of_birth', models.DateField(null=True)),
```
ericisaiah, line 34: I think we probably don't want to accept null values here.

app/futurevoter/models.py
```diff
@@ -0,0 +1,21 @@
...
+class FutureVoter(TrackingModel, UUIDModel, TimestampModel):
+    first_name = models.TextField(null=True)
+    last_name = models.TextField(null=True)
+    date_of_birth = models.DateField(null=True)
+    email = models.EmailField(null=True)
+    phone = PhoneNumberField(null=True, db_index=True)
+
+    class Meta:
+        ordering = ["-created_at"]
+        indexes = [
```
ericisaiah, line 16: Is there a reason phone is not in this list of indices, even though you added created_at? I may be missing something about Django

app/common/utils/serializers.py
```diff
@@ -104,4 +104,19 @@ def validate(self, data):
...
+
+    def dob_max_year(self):
+        return date.today().year - 16
```
ericisaiah, line 119: I see that was in the original, but why -16 here? Don't we want -12/13?

app/futurevoter/serializers.py
```diff
@@ -0,0 +1,66 @@
...
+            "embed_url",
+            "email_referrer",
+            "mobile_referrer",
+            "session_id",
+        ]
+
+    def dob_max_year(self):
+        # We do age validation below with a better error message, so just make
+        # sure the year isn't in the future
+        return date.today().year
```
ericisaiah, line 50: Why are you overriding this?

app/futurevoter/serializers.py
```diff
@@ -0,0 +1,66 @@
...
+from .models import FutureVoter
+
+
+class FutureVoterSerializer(
+    EnumSupportSerializerMixin, DateOfBirthSerializer, serializers.ModelSerializer
+):
+    first_name = serializers.CharField(required=True)
+    last_name = serializers.CharField(required=True)
+    date_of_birth = serializers.DateField(required=False)
+    email = serializers.EmailField(required=True)
```
ericisaiah, line 24: Do we need both email and phone? i.e. should we allow phone, email, or both? but not require both

app/futurevoter/serializers.py
```diff
@@ -0,0 +1,66 @@
...
+    zip_validator,
+)
+
+from .models import FutureVoter
+
+
+class FutureVoterSerializer(
+    EnumSupportSerializerMixin, DateOfBirthSerializer, serializers.ModelSerializer
+):
+    first_name = serializers.CharField(required=True)
```
ericisaiah, line 21: Is there a reason we require name?

ericisaiah (changes requested):
Various questions - some of which may be related to my missing context or Django knowledge. Hope some are helpful.

app/common/utils/serializers.py
benweissmann: So that subclasses (like FutureVoterSerializer) can override it

app/futurevoter/migrations/0001_initial.py
benweissmann: We always use nullable fields, because we run migrations before the code out so new columns have to be nullable or the old code will error. We *could* have non-nullable fields in brand new models like this one, but it feels weird to introduce that inconsistency (DB fields are non-nullable if they were in the first revision of the model, nullable otherwise). We enforce that this is required in the serializer rather than at the DB level

app/futurevoter/models.py
benweissmann: That's my bad; there shouldn't be an explicit indexes list here

app/futurevoter/serializers.py
benweissmann: Not sure, that's a Debra question

app/futurevoter/serializers.py
benweissmann: My next branch makes email optional; Debra wants to do all outreach by SMS so we'll need a phone but not an email

app/futurevoter/serializers.py
benweissmann: By default, you need to be 16 to use our tools. For this tool, you can be younger than that.

app/common/utils/serializers.py
ericisaiah: I thought we wanted to limit to 13 years or older

benweissmann: This is in `common/utils/serializers.py` -- it's what we use to serialized DOBs across the app. We override this in the subclass in `futurevoter/serializers.py` below.

app/futurevoter/migrations/0001_initial.py
ericisaiah: I would bet most young people at least stay in their same state, so that at least might be useful besides Zip.

app/futurevoter/models.py
benweissmann: Looked at this again -- I think this is correct. Django creates indexes for any field that has `db_index=True`, as well as any indexes listed here. The reason we don't set `db_index=True` on the `created_at` field in the parent `TimestampModel` is that some models want to set that index up differently (e.g. `Event` uses a composite index on `event_type, created_at`) and you can't set up a parent TimestampModel to allow subclasses to selectively disable a `db_index` (as far as i know).

app/futurevoter/serializers.py
ericisaiah: Shouldn't this be 13 years ago? Not just any year?

ericisaiah:
Just responded to a few things, but I don't think anything's a blocker. The only major thing I would make sure we do is add the state field. I can't imagine we wouldn't want to at least collect that location info. But maybe file a ticket to ask Debra when she's back versus making it a blocker.

benweissmann:
@ericisaiah https://app.clubhouse.io/turnout2020/story/5537/confirm-the-fields-we-want-for-futurevoter

app/futurevoter/serializers.py
benweissmann: We do a full validation below where we actually check birth date -- we can't validate the year in isolation, because we can't tell whether you're 12 or 13 based just on year number. So the validation below does the real over-18 validation, and this just makes it so a) the default validation doesn't kick in, and b) the error message to the client is "invalid year" if the year is in the future rather than "you must be 13"

PR #720, opened 2021-03-05
Update the Django admin to start making it more usable internally [#5362]

ericisaiah:
This PR includes
- Different UI-facing labels for models based on new schema (Organizations and Users)
- Adding a slightly more usable UI for Django admin (Suit) and add some proof of concept customizations
- Unregister (or don't register) certain models from the admin so they don't appear (for now, just comment out those we're no longer registering)
- Refactor `Accounts` app models to separate files for a proof of concept for separating concerns a bit more (I'd eventually like to do this for all the apps)
- Move time zones constant choices to a common lib


benweissmann:
This looks good! Big fan of splitting up the `models` file into separate files for each model, and the new admin theme looks great.

Instead of fully removing models from the django admin, is there a way to have them only show up for "blessed" users? Like we'd tag you/me as "sysadmins" or something, and those models would show up in the admin just for us. I think there's some models that we can remove entirely from the admin (like the 2fa models), but there's other ones (I'm thinking of Tasks in particular) that shouldn't be exposed to non-engineers but are really useful for debugging (for example, I added a feature a couple months back to allow us to re-run Tasks from the django admin when they fail, and that's been helpful a number of times).

Let me know if you want help digging into the test & lint failures that Travis is surfacing on this PR (for linting, you can run `make lint` in a separate terminal while `make up` is running to automatically reformat your code to match our lint rules; I configure my editor to format the code with `black` on save)

PR #722, opened 2021-03-09
FutureVoter welcome SMS

benweissmann:
- Add welcome SMS to FutureVoter
- Allow customizing the welcome message for a welcome sms
- Tracking FV welcomes separately from normal welcomes, because we specify a different message frequency in normal welcomes
- Make email optional for FutureVoter because apparently kids these days check their phones but not their emails??


app/futurevoter/api_views.py
```diff
@@ -11,3 +13,12 @@ class FutureVoterViewSet(CreateModelMixin, GenericViewSet):
...
     serializer_class = FutureVoterSerializer
     queryset = FutureVoter.objects.all()
+
+    def perform_create(self, serializer):
+        super().perform_create(serializer)
+
+        send_welcome_sms.delay(
+            str(serializer.validated_data["phone"]),
+            origin="futurevoter",
+            custom_message="Thanks for signing up to be a FutureVoter! We will send a text on your birthday to remind you to register to vote. Msg & Data rates may apply. Text STOP to unsubscribe.",
```
ericisaiah, line 23: I think the copy should say "on your 18th birthday"

app/futurevoter/serializers.py
```diff
@@ -21,7 +21,7 @@ class FutureVoterSerializer(
     first_name = serializers.CharField(required=True)
     last_name = serializers.CharField(required=True)
     date_of_birth = serializers.DateField(required=False)
-    email = serializers.EmailField(required=True)
+    email = serializers.EmailField(required=False)
```
ericisaiah, line 24: Is there a reason not to make first and last also required=False? The only thing we really need is phone # (and I would say add state, per my comment in the other PR)

app/smsbot/migrations/0012_number_futurevoter_welcome_time.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('smsbot', '0011_blast_new_fields'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='number',
+            name='futurevoter_welcome_time',
```
ericisaiah, line 15: Could you tell me more about why we need a FV-specific welcome time? Why not use the existing field? Would a phone number ever have both a `welcome_time` and `futurevoter_welcome_time`?

If we're trying to track the origin of a phone number, then I think a more scalable way to handle this is to make a new column that's `origin` that gets set to the origin you're already passing in, and that gets set to `VoteAmerica` by default (or other future apps).

ericisaiah (changes requested):

app/futurevoter/api_views.py
benweissmann: We're sending birthday texts every year, not just 18th (this language about how frequently we'll text you is required by our shortcode agreement with the carriers)

app/futurevoter/serializers.py
benweissmann: Let's flag this for Debra when we meet with her next; it's not a DB constraint so it's easy to change

app/smsbot/migrations/0012_number_futurevoter_welcome_time.py
benweissmann: It's because the welcome message for FVs ("we'll send you a text on your birthday...") is different than for regular tool users ("we send 2-4 txts/mo with election alerts...") -- our shortcode agreement with the carrier stipulates that we have to send a welcome message with opt-out instructions and message frequency, so we need to track this so we can re-welcome you if you sign up as a futurevoter (only getting 1 text per year) and then later you use the tools and get on our normal shortcode list (getting multiple texts per month).

ericisaiah: Right, I understand that part. My question is around adding it as an app-specific column. Why not make it a column that states the origin of the number instead of adding more columns that are specific only to one origin? Doing that would allow this functionality to scale more easily without having to add a new column every time there's a new origin app of phone numbers.

app/futurevoter/api_views.py
ericisaiah: Oh, I didn't realize we're texting on other years too. What's the purpose of reminding them on, say, their 15th birthday? Just curious, not criticism

benweissmann: I think debra wanted to a) keep in touch, and b) give them a link to our tool to pass onto anyone they know who's turning 18. Opened a ticket here to go over the copy and drip schedule with Debra: https://app.clubhouse.io/turnout2020/story/5538/confirm-copy-for-futurevoter-texts

app/smsbot/migrations/0012_number_futurevoter_welcome_time.py
benweissmann: In that scenario, would we have multiple `Number` records for a single phone number? We'd also have to switch the primary key to be a composite index on `(number, source)` rather than just using the phone number as a primary key, and be careful in the blast-messaging code to deduplicate based on phone number so we don't send double texts. My gut here is that having multiple records for a single number adds more complexity than we'd save by reducing the number of columns -- and I think that for the most part, we won't need to do this in very many cases (FutureVoter is a bit of a special case in terms of SMS opt-in; all of our other tools just use the standard opt-in language. We're going to want to get anyone over 18 onto the full SMS opted-in list so we can send election reminders and do fundraising; FutureVoter is likely the only place we'll interact with under-18s who don't need election reminders and probably aren't useful for fundraising)

ericisaiah: In my experience, it's a helpful polymorphic practice to scope records' uniqueness to an origin/category/etc so future features can avoid one-off columns and code blocks wherever they are referenced. This would also support putting app-specific code and message content in separate modules or overrides, and just include each one. While FutureVoter is the only other sms stream with different content right now, I can totally imagine there will be more in the future, like a stream for users coming in from colleges, maybe ones who registered while filing their taxes, etc. Anyway, may or may not be worth spending the effort at the moment.. Just something to consider to make this more scalable.

PR #724, opened 2021-03-15
Add birthday sms to FutureVoter

benweissmann:
Add pre-18th-birthday and 18th-birthday messages. Post-18th-birthday drip campaign will be a separate PR.


app/futurevoter/tasks.py
```diff
@@ -0,0 +1,94 @@
...
+def send_sms(number, message):
+    number.send_sms(message)
+
+
+@shared_task
+def birthday_under_18() -> None:
+    from .models import FutureVoter
+
+    today = date.today()
+
```
ericisaiah, line 30: I see what you did here, but I think you can greatly simplify this code with some fun django ORM lookups (or some manually SQL):

    from django.db.models import Q

    years = [(today.year - age) for age in range(13,18)]
    year_query = Q(date_of_birth__year__in=years)
    day_query = Q(date_of_birth__month=today.month, date_of_birth_day=today.day)
    day_query = day_query | Q(date_of_birth__month=2, date_of_birth_day=29) if today.month == 3 and today.day == 1
    voters = FutureVoter.objects.filter(year_query & day_query)

Something like that. Then I think you don't have to worry about the rest and constructing all the possible dobs for each year or having multiple if paths for leap years.

app/futurevoter/tasks.py
```diff
@@ -0,0 +1,94 @@
...
+    return (date(year, 3, 1) - timedelta(days=1)).day == 29
+
+
+# We extract this helper so we can mock it in tests
+def send_sms(number, message):
+    number.send_sms(message)
+
+
+@shared_task
+def birthday_under_18() -> None:
```
ericisaiah, line 26: I would combine the under/over birthday tasks, make one db call for all voters, then just do a switch on whether they should receive the general birthday sms or the 18th bday sms. While I'm not super concerned with the query performance given that this is a background task, it reduces the code maintainability significantly (1 task to maintain instead of 2) since you're doing the same thing for each voter and just changing the message (right? or did i miss something?)

ericisaiah (changes requested):

app/futurevoter/tasks.py
```diff
@@ -0,0 +1,94 @@
...
+
+    if today.month == 2 and today.day == 29:
+        # It's a leap day -- nobody turns 18 on leap days. Return early so
+        # we don't try to call date() with feb 29 on a non-leap year, because
+        # that will throw
+        return
+
+    for futurevoter in FutureVoter.objects.filter(
+        date_of_birth__lte=date(today.year - 18, today.month, today.day),
+        last_drip_message__isnull=True,
```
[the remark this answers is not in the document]
benweissmann: They're the same in most but not all cases -- for `jsonb` fields, `= None` means that JSON `null` is stored in the field, whereas `__isnull` means the field itself is a postgres `null` (see: https://stackoverflow.com/a/59313851).

I think we generally use `__isnull` through the codebase but I don't have a strong preference in either direction.

app/futurevoter/models.py
```diff
@@ -7,9 +9,10 @@
 class FutureVoter(TrackingModel, UUIDModel, TimestampModel):
     first_name = models.TextField(null=True)
     last_name = models.TextField(null=True)
-    date_of_birth = models.DateField(null=True)
+    date_of_birth = models.DateField(null=True, db_index=True)
     email = models.EmailField(null=True)
     phone = PhoneNumberField(null=True, db_index=True)
+    last_drip_message = models.DateTimeField(null=True, db_index=True)
```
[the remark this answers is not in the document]
benweissmann: It's going to be used for the post-18th-birthday drip messages -- we'll be texting these folks once a week (for up to 5 weeks) until they confirm they're registered, so this will track the last time we sent one of those messages (we set it when we send the 18th bday message so that we will be able to query for `last_drip_message < today() - '7 days'`)

app/futurevoter/tasks.py
benweissmann: The big difference is how we select the voters. For under-18 birthday texts, we select voters whose birthday is today. For 18th birthday texts, we select voters who are over 18 and haven't gotten the text yet -- the difference is in error handling. If the daily birthday job fails, we don't want to send a late birthday text to under-18s, but we do want to keep trying to text folks when they turn 18 if the 18th-birthday job fails.

ericisaiah: > For 18th birthday texts, we select voters who are over 18 and haven't gotten the text yet
Isn't that the same thing as checking each day to see if someone turned 18? I'm not sure i understand the difference.

app/futurevoter/models.py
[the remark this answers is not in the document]
ericisaiah: Ah, I see. Do you mean that additional 5-week drip-message feature is forthcoming?

app/futurevoter/tasks.py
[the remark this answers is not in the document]
ericisaiah: I did look that up, but was trying to understand if there is a reason you used it since this isn't a jsonb field. I was just curious since it seemed odd to me, but no big deal. As long as it's readable, which it is.

app/futurevoter/tasks.py
benweissmann: It's different if the job fails for whatever reason (db downtime, server crash, etc).

For the under-18 birthday texts, if the job fails partway through, we just let it fail. It's not a big deal if we don't send out some under-18 birthday texts, and it'd look bad if we sent those a day late.

For the 18th birthday text, if the job fails partway through, we won't get all the birthday texts out that day. But the next day, because we're selecting all users over 18 who haven't yet received the text *even if it's not still their birthday*, we'll pick up those users that we failed to send a text to yesterday, and send them the text (albeit a day late) -- I think that's desirable, because we really don't want to miss sending the "you're 18, time to register" text, even if it's late.

app/futurevoter/models.py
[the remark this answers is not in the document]
benweissmann: Yup, i'm planning to do that as a follow-up PR: https://app.clubhouse.io/turnout2020/story/5520/futurevoter-post-18th-birthday-drip-campaign

app/futurevoter/tasks.py
ericisaiah: Given those differing needs, makes sense to keep them separate then!

app/futurevoter/tasks.py
benweissmann: yeah, that looks great! The only tweak is that we should only `|` in feb 29 on non-leap-years -- if we do that on leap years, then feb 29 birthdays will get texted twice (once on feb 29 and once on march 1).

benweissmann:
@ericisaiah can you take another look? Couple changes:

- Reworked the leap year handling like you suggested
- Added a mandatory zipcode field and make last name optional
- Made futurevoters unique on phone number, so we don't send double texts if someone fills out the form twice

app/futurevoter/tasks.py
```diff
@@ -0,0 +1,81 @@
...
+
+    today = date.today()
+
+    years = [(today.year - age) for age in range(13, 18)]
+    year_query = Q(date_of_birth__year__in=years)
+    day_query = Q(date_of_birth__month=today.month, date_of_birth__day=today.day)
+    if today.month == 3 and today.day == 1 and not is_leap_year(today.year):
+        # Include Feb 29 birthdays on March 1
+        day_query = day_query | Q(date_of_birth__month=2, date_of_birth__day=29)
+    FutureVoter.objects.filter(year_query & day_query)
```
ericisaiah, line 39: Did you mean to keep this line here?

app/futurevoter/tasks.py
```diff
@@ -0,0 +1,81 @@
...
+        # we don't try to call date() with feb 29 on a non-leap year, because
+        # that will throw
+        return
+
+    for futurevoter in FutureVoter.objects.filter(
+        date_of_birth__lte=date(today.year - 18, today.month, today.day),
+        last_drip_message__isnull=True,
+    ).iterator():
+        try:
+            phone = Number.objects.get(pk=str(futurevoter.phone))
```
ericisaiah, line 68: I didn't see this before, but I think you should call a `phone_number` related object on FutureVoter, so you can just do `send_sms(futurevoter.phone_number...`. and do the lookup in the model.

app/futurevoter/tasks.py
```diff
@@ -0,0 +1,81 @@
...
+    year_query = Q(date_of_birth__year__in=years)
+    day_query = Q(date_of_birth__month=today.month, date_of_birth__day=today.day)
+    if today.month == 3 and today.day == 1 and not is_leap_year(today.year):
+        # Include Feb 29 birthdays on March 1
+        day_query = day_query | Q(date_of_birth__month=2, date_of_birth__day=29)
+    FutureVoter.objects.filter(year_query & day_query)
+
+    for futurevoter in FutureVoter.objects.filter(year_query & day_query).iterator():
+        try:
+            phone = Number.objects.get(pk=str(futurevoter.phone))
```
ericisaiah, line 43: same comment below

ericisaiah (changes requested):
Couple of minor things, maybe a stray line

ericisaiah (commented):
Love it!

PR #725, opened 2021-03-16
Fix bug in voter lookup background task

benweissmann:
Context: we have a background task (fired off when someone completes an action on our site, like registering to vote) that looks the person up with Targetsmart and stores the info for later analysis. We have bits of a system that goes along with this to re-check periodically so we know when our users appear/disappear from the voter rolls, etc, but we're not currently using that.

When I reworked the Verify algorithm, I changed the signature of `query_targetsmart`, which broke this background task. This isn't super urgent because it's just used for internal data collection right now, but good to resolve anyway.

I didn't catch this in development because these voter lookup tasks happen in a separate queue (so they don't block more important tasks) and we hadn't configured the development Celery worker to read from the `voter` queue.


ericisaiah:
I'm not sure I have enough context to be helpful on this one, but I trust you've done what you need to do here. The only suggestion might be to identify any useful test for the arg signature, but may not be worth it if this isn't that important.

ericisaiah (commented):
I'm not sure I have enough context to be helpful on this one, but I trust you've done what you need to do here. The only suggestion might be to identify any useful test for the arg signature, but may not be worth it if this isn't that important.

benweissmann:
We do have tests that cover this function -- I just updated the code and those tests but didn't spot this call site 🤦

PR #726, opened 2021-03-24
Initial admin reports through Django admin. [#5584]

ericisaiah:
https://app.clubhouse.io/turnout2020/story/5584/proof-of-concept-standalone-basic-reports-in-django-admin


benweissmann (approved):
Looks good! Couple notes:

- looks like you have some build failures from mypy
- Not necessary for this POC, but long-term we should do any reporting queries against the [readonly DB](https://github.com/vote/turnout/blob/ac964a960237e1889e885214e30fee105d61d2f9/app/turnout/settings/core.py#L48) -- see [this example](https://github.com/vote/turnout/blob/365a7515a667297b7980832e00e5e3710a97d05f/app/smsbot/models.py#L130). In prod, we use Amazon Aurora which runs read-only replicas for failover, and using the readonly database connection will offload the query load to those replica nodes rather than increasing load on the primary (but be aware that the replication is async, though usually <1 second delayed)

ericisaiah:
It looks like mypy is failing because it's incorrectly flagging a missing method. I'm inheriting from two classes, and it seems to be missing the second one (BaseReport). I'm not familiar enough with mypy to know if this is something we can tweak in settings. A quick google says at least mypy version 0.812 is supposed to support multiple inheritance. Not sure about our version 0.770. This check seems a little unnecessary to me (nothing about this route would work if the method were missing), so I would rather just ignore it. Any other thoughts?

ericisaiah:
Actually, that's not right. I forgot each report class is overriding that method anyway, so the views (e.g. VotersByStateReportView) are implementing the method directly. Do you know why it thinks it's missing an attribute? If not, I'll just ignore it.

PR #732, opened 2021-04-08
Show a disabled message for users whose own account or organization is disabled

ericisaiah:
Also some tests.


ericisaiah:
https://app.clubhouse.io/turnout2020/story/5385/subscribers-make-sure-that-every-user-is-associated-with-a-client

app/accounts/models/user.py
```diff
@@ -123,3 +123,12 @@ def note_access(self, ip=None) -> Access:
             self.active_client.latest_access = access
             self.active_client.save(update_fields=["latest_access"])
         return access
+
+    def is_organization_owner(self):
```
benweissmann, line 127: Is this just a short-term thing? I think if we want to really do org owners, we should move to having a foreign key from organization to its owner

benweissmann: But i also think this is fine for now

app/accounts/templates/accounts/disabled_account.html
```diff
@@ -0,0 +1,25 @@
...
+        <div class="row">
+            <div class="col-md-6 offset-md-3 public-page-wrap">
+                <div class="row">
+                    <h2>Your account is deactivated</h2>
+                    <p>
+                        This account has been deactivated.
+                        {% if is_organization_owner %}
+                            Please contact <a href="mailto:sales@voteamerica.com">sales@voteamerica.com</a> to reactivate your account.
+                        {% else %}
+                            Please contact your team admin to reactivate your account.
```
benweissmann, line 16: Can we show the billing email here so they know who the account owner is?

app/manage/mixins.py
```diff
@@ -5,6 +5,12 @@
 
 class ManageViewMixin(LoginRequiredMixin):
     def dispatch(self, request, *args, **kwargs):
+
+        if request.user.is_authenticated and (
```
benweissmann, line 9: Does this cover the VoteAmerica admin in addition to the Django admin? Not sure if ManageViewMixin is included everywhere or just the django admin

benweissmann (changes requested):

app/accounts/models/user.py
ericisaiah: Yeah, I am intentionally avoiding a migration since clients/partners/subscribers/orgs need to refactored in any case and rethought in the near term...

app/accounts/templates/accounts/disabled_account.html
ericisaiah: Sounds good

app/manage/mixins.py
ericisaiah: Yeah, that's what the test `test_disabled_account` confirms. Can you double check that?

benweissmann (approved):
Looks good

ericisaiah:
@benweissmann Could you take another look at this please? The changes to get this working were not obvious and required some digging into how two_factor overrides the built-in Django AuthenticationForm. You'll see that comment in-line, but let me know if you have any questions. We're basically doing something different than Django or two_factor expect with `User.is_active`.

Also I've added the UI to actually deactivate/reactive users. After this gets merged and deployed, I'll finally fix those orphan users so they are re-associated with their clients.

benweissmann (approved):
looks good to me!

ericisaiah:
thanks!

PR #733, opened 2021-04-13
add blackout and brownout feature flags to the public election information api

benweissmann:
This adds two new Optimizely feature flags:

- `state_info_brownout` causes us to append a message to the text of all API fields, providing a visible warnings that the API is going away
- `state_info_blackout` replaces all the state info with a message saying that the API is no longer available

Each of these has a customizable `message` that can be controlled from Optimizely.

https://app.clubhouse.io/turnout2020/story/5504/turn-off-unauthenticated-election-api-access


ericisaiah (approved):
I don't know enough about the feature flag tool to have an informed opinion on how to do this correctly, but looks reasonable. This looks like a generalized way to brownout/blackout fields overall, but does this mean we have to add feature flags for every state info field type?

benweissmann:
No, the blackout/brownout fields are global. You set them once and they apply to every field. So there's just two flags: `state_info_brownout` and `state_info_blackout`. They have a customizable "message" property that lets you override the message, but that's still global across all fields.

ericisaiah:
great!

PR #734, opened 2021-04-16
Add more instructions to registration forms

benweissmann:
Backend support for https://github.com/vote/voteamerica/pull/1188

This adds additional instructions that used to be provided on the frontend to the cover sheet and email for Register.


app/register/generateform.py
```diff
@@ -48,6 +48,22 @@
     "register/templates/pdf/blank-register-forms-cover-8pt.pdf"
 )
 
+INSTRUCTIONS = """
+1. Print and complete this document. If you don’t have a printer, your local library will print it for you.
+
+2. Fill in box 6 (ID Number). %s
```
ericisaiah, line 54: Will this insert the ID here? Minor thing but I would make this a colon versus period.

app/register/generateform.py
```diff
@@ -48,6 +48,22 @@
...
+
+3. Box 8 (Race or Ethnic Group) is optional.
+
+4. Read over the form and fill out any additional boxes that apply to you.
+
+5. Sign and date the form. This is very important.
+
+6. Mail or hand-deliver the form to the address listed below.
```
ericisaiah, line 62: Maybe, "Send by mail or deliver the form in person to..."

app/register/generateform.py
```diff
@@ -48,6 +48,22 @@
...
+
+7. If you need a stamp, you can buy one at any grocery store. Just ask the cashier.
```
ericisaiah, line 64: This seems like it could be a parenthetical as a a part of step 6 (and the fewer steps the better). Perhaps also being clear that a stamp is required to mail it because sometimes people get confused if they need a stamp to mail voter stuff (even though i know in this case it's just something they're printing out). Another small thing, but just to be more inclusive: I might use 'convenience' store or something more general/common vs grocery store. Interesting research out there on distance to travel for groceries in particular: e.g. https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3701025/

So I would add something like this to step 6: "(You'll need a stamp to mail it; you can buy one at your local convenience store.)"

app/register/templates/register/email/file_notification.html
```diff
@@ -35,6 +35,9 @@
   <li>Click the button above to download your form</li>
   <li>Check for the PDF file in the "Downloads" folder on your mobile device or computer and then open it</li>
   <li>Print your form</li>
+  <li>Fill in box 6 (ID Number). {{ state_info.registration_nvrf_box_6 }}</li>
+  <li>Box 8 is optional.</li>
```
ericisaiah, line 39: Add the parenthetical here to be consistent (Race or Ethnic Group)?

app/register/templates/register/email/file_notification.html
```diff
@@ -35,6 +35,9 @@
   <li>Click the button above to download your form</li>
   <li>Check for the PDF file in the "Downloads" folder on your mobile device or computer and then open it</li>
   <li>Print your form</li>
+  <li>Fill in box 6 (ID Number). {{ state_info.registration_nvrf_box_6 }}</li>
```
ericisaiah, line 38: same commont, colon vs period

ericisaiah (changes requested):
A few suggestions

app/register/generateform.py
benweissmann: Nope, we insert the value of `registration_nvrf_box_6` here. We store this as a full sentence; see https://admin.voteamerica.com/manage/election/fieldinformationtype/registration_nvrf_box_6/

app/register/templates/register/email/file_notification.html
benweissmann: Let's run through this copy real quick with Debra today (have some other questions about these instructions too)

app/register/generateform.py
ericisaiah: Ohhh gotcha. Nevermind then.

app/register/templates/register/email/file_notification.html
ericisaiah: ignore

ericisaiah (approved):
looks good

PR #737, opened 2021-04-28
Manage customers, subscriptions, and users in Django.

ericisaiah:


ericisaiah:
https://app.clubhouse.io/turnout2020/story/5672/django-admin-manage-subscribers-and-their-various-settings-https-admin-voteamerica-com-manage-subscription-subscribers

PR #743, opened 2021-05-04
Bring back the ability to edit state information form field types.

ericisaiah:
These objects can be optionally read-only for certain staff using Django's built-in permissions.


ericisaiah:
https://app.clubhouse.io/turnout2020/story/5643/add-the-ability-to-edit-an-api-field-back-into-django

PR #744, opened 2021-05-06
Action log browser and basic lookup functionality across tools.

ericisaiah:


ericisaiah:
https://app.clubhouse.io/turnout2020/story/5616/action-event-log-inside-of-django-admin

ericisaiah:
@benweissmann Could you review this one to look at the SQL? I want to make sure I'm correctly pulling in from the right tables and matching up actions. I believe I am based on my local testing, but want to verify.

app/turnout/admin/views.py
```diff
@@ -1,4 +1,109 @@
...
+    def search_log_and_set_context(self, context, query):
+
+        page = 1
+        page_size = 100
+        search_terms = []
+
+        sql_count = "SELECT COUNT(*) "
+        sql_fetch = "SELECT * "
+
+        sql = """
```
ericisaiah, line 26: @benweissmann Specifically this area

app/templates/admin/action_logs.html
```diff
@@ -0,0 +1,59 @@
...
+
+    <table class="table table-striped table-report">
+        <thead>
+            <th>Action</th>
+            <th>Customer Source</th>
+            <th>First name</th>
+            <th>Last name</th>
+            <th>City</th>
+            <th>State</th>
+            <th>Timestamp</th>
```
benweissmann, line 37: We had a convention in the non-Django admin of always putting the user's timezone in the header if we were localizing the the user's timezone (e.g. `Timestamp (US/Eastern)`); might be a good thing to continue in the Django admin

app/templates/admin/action_logs.html
```diff
@@ -0,0 +1,59 @@
...
+        </thead>
+        <tbody>
+            {% for action in actions %}
+            <tr>
+                <td>{{ action.tool }}</th>
+                <td><a href="{% url 'action_logs' %}?customer_name={{action.customer_name}}&customer_id={{action.customer_id}}">{{ action.customer_name }}</th>
```
benweissmann, line 43: These user-controlled values should get passed through `|urlencode` (I'm not sure we do any validation on customer name, so it could include a `"` and mess up this HTML)

app/turnout/admin/views.py
```diff
@@ -1,4 +1,109 @@
...
+            FROM (
+                (
+                    SELECT 'Absentee Ballot Request' AS tool, first_name, last_name, actions.email, phone, state_id AS state, city, action_id, actions.created_at, partner_id AS customer_id, customers.name AS customer_name
```
benweissmann, line 29: I think we generally just refer to this tool at `Absentee` so that's probably what Debra would expect here, but whatever makes the most sense

benweissmann (approved):
Looks great, just a couple flags

app/templates/admin/action_logs.html
ericisaiah: Yeah, that makes more sense rather than just repeating the same timezone over and over in the cell rows. Will do.

app/templates/admin/action_logs.html
ericisaiah: Definitely a good call. However, this manual process (including my original version) really feels icky to me overall. I'm having a hard time believing Django's url template tag can't handle this already as extra arguments, but after looking closely at the documentation and searching around, it seems no one has a good answer except to do this basically (or create a new method that does this for me, which is overkill). Anything you think I'm missing? Otherwise I'll just proceed with this janky approach.

app/turnout/admin/views.py
ericisaiah: Good to know, thanks!

app/templates/admin/action_logs.html
benweissmann: Yeah, i'm not aware of anyway to assemble query parameters other than this :(

ericisaiah: I can confirm the url template tag doesn't handle this: https://github.com/django/django/blob/stable/2.2.x/django/template/defaulttags.py#L1315 So strange...!

PR #757, opened 2021-06-22
Use private endpoint for importfromprod

benweissmann:
Switch the importfromprod command to use the new private data endpoint.


ericisaiah:
Could you share more about the difference between `/state` and `/state/private`? It looks like one shows a warning and one doesn't?

benweissmann:
`/state/` is the current endpoint for the soon-to-be-discontinued public API. `/state/private/` is the same endpoint, but at a new URL and with a warning that the data can only be used on voteamerica.com. `/state/private/` is what the frontend / website is using -- it's the exact same thing as the old `/state/` API, but it won't be turned off when we enable the brownout/blackout flags. More details in this clubhouse ticket: https://app.clubhouse.io/turnout2020/story/5430/authenticated-state-information-api

ericisaiah:
Ah yes, thank you. I was having a mental block forgetting the connection with that work. Got it. Looks good. I'm going to push a change to this branch to update the old premium language, but otherwise will merge.

ericisaiah (approved):

PR #759, opened 2021-07-10
Eliminate interests and subscriptions. Allow customers to sign up directly and enable management of them in Django Admin.

ericisaiah:
In this PR:

- Remove management of Subscriptions and Interests in the old front-end, and instead make it available only in Django.
- Refactor multi_tenant models to be separate files, which makes them more maintainable and reduces circular import hell. (Some removal of imports is due to this issue.)
- Move single-purpose client-related tasks to the Client model - removes the multi_tenant/invite.py file. I'm not sure why it was separate.
- Move subscription information to clients (customers). Keep the Subscription model for now so that the data migration can run.
- Have the subscribe form create clients (customers) directly, which then go into a "pending" state. Then VA admins can set a customer to "active," which will initiate the setup functions that activating an Interest did previously.
- Some outstanding migration changes, which do not affect the database tables, are included in the auto-generated migrations.


app/common/enums.py
```diff
@@ -297,11 +299,13 @@ class SubscriberPlan(Enum, metaclass=EnumMeta):
     FREE = "free"
     PREMIUM = "premium"
     NONPROFIT = "nonprofit"
+    SYSTEM = "system"
```
ericisaiah, line 302: This is forward-looking, but doesn't do anything yet. I'd like to replace the `Client.objects.first()` references with e.g. `Client.objects.get(plan= SubscriberPlan.SYSTEM)` so it's both easier to understand what's happening and it won't break if the Client instance ever changes.

app/verifier/tests/test_api_views.py
```diff
@@ -338,10 +342,33 @@ def test_inactive_subscriber(mock_apicalls, mock_finish):
         "last_update_date": "2021-01-21",
     }
 
+    lookup = Lookup.objects.first()
+
+    # Subscribers are now "pending" by default, which has the same effect as "disabled"
```
benweissmann, line 347: I'd suggest also adding tests to verify that "pending" subscribers:

- Can't hit the election information API
- Have their records (`Lookup`, `Registration`, etc.) flagged with `visible_to_subscriber: False` (this flag will prevent the data from being included in ActionNetwork syncs, Bluelink/TMC syncs, and customer-facing exports and lists)

benweissmann (approved):
Looks good! Just a few comments but overall looks great.

app/accounts/models/user.py
```diff
@@ -94,7 +93,7 @@ def active_client_slug(self):
     @cached_property
     def allowed_clients(self):
         if self.can_manage_subscribers:
-            return Client.objects.all()
+            return self.active_client.__class__.objects.all()
```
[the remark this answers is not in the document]
ericisaiah: yep. I think this needs a slightly bigger refactor but I was trying to keep changes germane.

app/multi_tenant/models/client.py
```diff
@@ -0,0 +1,364 @@
...
+        from multi_tenant.models import SubscriberSlug
+
+        if not self._state.adding:
+            # Get database values for comparison
+            existing_customer = Client.objects.get(pk=self.pk)
+
+            # Set the customer's slug if one is missing
+            if not self.default_slug:
+                subscriber_slug = SubscriberSlug.objects.create(
+                    slug=slugify(self.name), subscriber=self
```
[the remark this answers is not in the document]
ericisaiah: Ah thank you for that flag. I casually assumed `slugify` already did that checking. I didn't realize it was just literally filtering the string. Given the rarity of this issue, I will probably just make this a separate ticket and implement something like this instead: https://pypi.org/project/awesome-slugify/

app/verifier/tests/test_api_views.py
ericisaiah: Great idea. I will add those.

